PR #64, opened 2016-01-08
Copy-pasteable Glide instructions

bspaans:
My shell scripts bring all the boys to the yard


rusenask:
Thanks @bspaans for your copy-pastable instructions!

PR #101, opened 2016-02-10
Change API access URIs to be more convenient (for me!)

daniel-bryant-uk:
Hey Karolis,

These are just suggestions, but I've replaced proxy_host with localhost, as I found this easier when cutting and pasting. I've also added a couple of links too - just for convenience :-)


rusenask:
Hey Daniel,
Thanks!

PR #161, opened 2016-03-18
Feature/compact json xml req bodies

rusenask:
compacting json/xml request bodies before matching


import.go
```diff
@@ -116,14 +117,29 @@ func (d *DBClient) ImportFromURL(url string) error {
...
+
 // ImportPayloads - a function to save given payloads into the database.
 func (d *DBClient) ImportPayloads(payloads []Payload) error {
 	if len(payloads) > 0 {
 		success := 0
 		failed := 0
 		for _, pl := range payloads {
+			pl.Request.Headers = make(map[string][]string)
+			// sniffing content types
+			if isJSON(pl.Request.Body) {
```
mogronalol: Could you not check for `JSON` by checking if the content type header is `JSON`?

rusenask: so I as we spoke earlier - it just adds missing headers by sniffing content. Updated it to ignore this step if headers are there.

PR #165, opened 2016-03-23
Feature/custom certs

rusenask:
added functionality:
* override default certificate for MITM. 
* generate your own certificates with Hoverfly.


cmd/hoverfly/main.go
```diff
@@ -106,6 +118,45 @@ func main() {
 		log.SetFormatter(&log.TextFormatter{})
 	}
 
+	if *generateCA {
```
mogronalol: I think this main method is getting quite long, is there a way to split things up a bit more?

mogronalol: For example could the certificate stuff go in it's own package?

PR #219, opened 2016-07-18
Merge in hoverctl logs, hoverfly cache rebuilding and hoverfly partial matching on URL and HTTP Method

benjih:


benjvi:
LGTM ;)

PR #226, opened 2016-07-26
Partial matching first pass

benjvi:
Added request template entity which allows for matching on only the parts of the request you wan t (out of host, path, method, query, body, headers). Only headers you specify will be matched on

Using `hoverctl templates $FILEPATH` to import additional templates and `hoverctl templates` to show request template data loaded as the native JSON. Use `hoverctl delete templates` to delete. These call the /api/templates endpoint in hoverfly with POST, GET and DELETE methods respectively

The request templates are used as a fallback in case an exact match is not found in the request cache. The first template to match will be the one to return its response

This doesn't cover any wildcard or regex matching in fields such as the path and body


benjih:
lgtm

PR #236, opened 2016-08-02
Gofmt

benjih:
I created a target in the Makefile that will go fmt all the Hoverfly source files.

And then I ran it.


daniel-bryant-uk:
LGTM!

PR #237, opened 2016-08-03
Changes to hoverctl and documentation regarding delays

benjih:


tjcunliffe:
@benjih the readme says`This example will add a 2 second delay to responses to github.com using any HTTP method`, but in the `delays.json` file it says `"urlPattern": "host1"`. Am I missing something?

benjih:
Nope, I wrote the README with what the example JSON file should be but didn't make the change to reflect the description.

PR #243, opened 2016-08-17
Headers in webserver mode

benjih:
@paprende discovered a bug with webserver mode.

When using webserver mode, the responses that it serve don't have the headers that are stored in the cache.

This branch fixes this by iterating through the headers on the generated response and then adds them to the response being written in the webserver.


daniel-bryant-uk:
LGTM!

PR #256, opened 2016-09-01
RequestTemplates now allow for * wildcards

benjih:


core/matching/request_template_store.go
```diff
@@ -99,18 +99,26 @@ func (this *RequestTemplateStore) Wipe() {
...
+		requestTemplateValues, ok := requestHeaders[templateHeaderKey]
+		if !ok {
 			return false
 		}
+
+		for _, templateHeaderValue := range templateHeaderValues {
+			found := 0
+			for _, requestHeaderValue := range requestTemplateValues {
+				if glob.Glob(templateHeaderValue, requestHeaderValue) {
+					found++
```
daniel-bryant-uk: Is it important to count the finds, or can we short-cut the return here?

core/matching/request_template_store_test.go
```diff
@@ -373,7 +373,7 @@ func TestRequestTemplateResponsePairCanBeConvertedToARequestResponsePairView_Whi
 			Method: &method,
 		},
 		Response: models.ResponseDetails{
-			Body: "Yo",
+			Body: "template matched",
```
daniel-bryant-uk: Haha, you've gone all enterprise! :-)

core/matching/request_template_store_test.go
```diff
@@ -386,5 +386,220 @@ func TestRequestTemplateResponsePairCanBeConvertedToARequestResponsePairView_Whi
...
+
+	Expect(response.Body).To(Equal("template matched"))
+}
+
+func TestTemplatesCanUseGlobsOnSchemeAndBeMatched(t *testing.T) {
+	RegisterTestingT(t)
+
+	requestTemplateResponsePair := RequestTemplateResponsePair{
+		RequestTemplate: RequestTemplate{
+			Scheme: StringToPointer("H*"),
```
daniel-bryant-uk: From this test it appears that the globbing is case-insensitive, which would be find for the method etc, but could it cause any edge cases? (and does this violate the principle of least surprise e.g. with grep I have to turn on case-insensitivity)?

daniel-bryant-uk:
A couple of quick questions @benjih, but other than that LGTM!

core/matching/request_template_store_test.go
benjih: This is an interesting one. The globbing library used tries to match how UNIX handles wildcards. In UNIX, the globbing is only on file structures where the file names are case insensitive.

core/matching/request_template_store.go
benjih: Good question.

I tried that and some tests failed which then made me remember why it is done like this.

In Hoverfly, the value of a header is an array of strings. There for, your template could potentially have multiple values to match on. Short-cutting the return when stop this functionality from working as you would return on your first match.

Your template:
```
Content-Type: [text/plain, application/json]
```

Unmatching request:
```
Content-Type: [text/plain]
```

benjih:
While looking at the tests to make sure the count in the header matching was required, I found a test that was not being run and was also unimplemented, so I've added it and implemented.

Also added some more tests while I was there.

PR #276, opened 2016-10-03
Tidy up and removing views from ResponseDelays

benjih:
For sanity's sake, I moved all of the new interface implementations which are used by the handlers into their own file called hoverfly_services.go. This isn't permanent, but we need to trim down Hoverfly before I can move them in again.

Also, removed the view code form ResponseDelay so now they are clearly separated so that I can ensure that the views for the /v1 endpoints don't change.


daniel-bryant-uk:
LGTM!

PR #278, opened 2016-10-14
Created /api/v2/simulation

benjih:


core/models/payload.go
```diff
@@ -295,3 +314,31 @@ func (r *ResponseDetails) ConvertToResponseDetailsView() v1.ResponseDetailsView
...
+	// Check headers for gzip
+	contentEncodingValues := r.Headers["Content-Encoding"]
+	if len(contentEncodingValues) > 0 {
+		needsEncoding = true
+	} else {
+		mimeType := http.DetectContentType([]byte(r.Body))
+		needsEncoding = true
+		for _, v := range supportedMimeTypes {
+			if strings.Contains(mimeType, v) {
+				needsEncoding = false
```
danielbryantuk: This code block is a little unclear (it might be the variable name, or perhaps two variables needed - needsEncoding and canEncode ?)

functional-tests/core/ft_api_test.go
```diff
@@ -27,15 +28,67 @@ var _ = Describe("Interacting with the API", func() {
...
+		
+		BeforeEach(func() {
+			ImportHoverflyRecords(jsonRequestResponsePair1)
+			ImportHoverflyRecords(jsonRequestResponsePair2)
+			SetHoverflyResponseDelays("testdata/delays.json")
+		})
 
-		It("Should get all the Hoverfly config available", func() {
-			req := sling.New().Get(hoverflyAdminUrl + "/api/v2/hoverfly")
+		It("Should get all the Hoverfly simulation data in one JSON file", func() {
```
danielbryantuk: This function is quite a beast, but I'm guessing this is simply due to the style imposed by Golang? e.g. the repeated "Expect(err).To(BeNil())"

danielbryantuk: Thanks Benji - completely makes sense (I'm just longing for Java/JUnit semantics ;-) )

core/models/payload.go
benjih: I agree with you @danielbryantuk. Having gone over it myself, its not as readable as it could be.

I've added comments explaining the functionality of that code to both of the places it occurs. Once we strip out the v1 API code from Hoverfly, I'll refactor this.

PR #283, opened 2016-10-25
Created and started using an interface for requests, responses and pairs for converting views into model structs

benjih:


danielbryantuk:
LGTM!

PR #288, opened 2016-11-01
Duplicated a load of tests for the new ConvertTo functions

benjih:
These tests are pretty much a duplicate of the ones being used to test ConvertToV1*. The functionality of these functions are the same but return either a v1 or v2 view struct.


danielbryantuk:
LGTM

PR #289, opened 2016-11-02
Templates are now included in the GetSimulation

benjih:
Beforehand, I had only included recordings into the new simulation endpoint. The reason for this is that it required more refactoring to include in the templates in a nice way.


danielbryantuk:
LGTM!

PR #291, opened 2016-11-09
Removing push, pull and the localcache from hoverctl, updating import and export to work with paths

benjih:


hoverctl/file.go
```diff
@@ -0,0 +1,26 @@
...
+
+	log "github.com/Sirupsen/logrus"
+)
+
+func WriteFile(filePath string, data []byte) error {
+	basePath := path.Dir(filePath)
+	fileName := path.Base(filePath)
+	log.Debug(basePath)
+
+	err := os.MkdirAll(basePath, 0777)
```
danielbryantuk: Minor one, but should we being using 777 for the directory (we're using 644 for the file below)?

benjih: No. We shouldn't.

PR #295, opened 2016-11-10
Errors on the put method for v2 simulations now error properly

benjih:
and the empty bucket error should no longer trigger an error on deleting.


danielbryantuk:
LGTM!

PR #303, opened 2016-11-18
The destination matches on the host and path

benjih:
There's also 

`hoverctl destination '<destination-regex>'

and

`hoverctl destination '<destination-regex>' --dry-run url-to-test.com


danielbryantuk:
LGTM!

PR #309, opened 2016-11-25
Added a config command to get the configuration being used by hoverctl as well as the location of the config file

benjih:


danielbryantuk:
LGTMA!

PR #313, opened 2016-11-29
Fixed a bug where Hoverfly blocks after failing to find a request in the memory_cache

benjih:
Added a unit and functional test around this.


danielbryantuk:
LGTM!

PR #315, opened 2016-12-01
Updated the error messages in the v2 simulation handler

benjih:
And now log the actual errors to debug.

### DIFF
--- a/core/handlers/v2/simulation_handler.go
+++ b/core/handlers/v2/simulation_handler.go
@@ -64,7 +64,7 @@ func (this *SimulationHandler) Put(w http.ResponseWriter, req *http.Request, nex
 			"body": string(body),
 		}).Debug(err.Error())
 
-		handlers.WriteErrorResponse(w, "Could not import simulation, did not follow valid v2 schema", http.StatusInternalServerError)
+		handlers.WriteErrorResponse(w, "Could not import simulation, was not valid json", http.StatusInternalServerError)
 		return
 	}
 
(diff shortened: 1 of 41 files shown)


core/handlers/v2/simulation_handler.go
```diff
@@ -57,15 +59,25 @@ func (this *SimulationHandler) Put(w http.ResponseWriter, req *http.Request, nex
 
 	err := json.Unmarshal(body, &simulationView)
 	if err != nil {
-		handlers.WriteErrorResponse(w, err.Error(), http.StatusInternalServerError)
+
+		log.WithFields(log.Fields{
+			"body": string(body),
+		}).Debug(err.Error())
+
+		handlers.WriteErrorResponse(w, "Could not import simulation, did not follow valid v2 schema", http.StatusInternalServerError)
```
benjih: So it was kinda doing this, but I wasn't outputting the right error messages. Fixed up so now it will report if the schema was wrong or if the JSON was invalid. Changed a test and added a new test to these both these errors.

core/handlers/v2/simulation_handler.go
```diff
@@ -53,19 +55,40 @@ func (this *SimulationHandler) Get(w http.ResponseWriter, req *http.Request, nex
...
 
-	err := json.Unmarshal(body, &simulationView)
-	if err != nil {
-		handlers.WriteErrorResponse(w, err.Error(), http.StatusInternalServerError)
+	if path, err := simulationView.GetValidationSchema().Validate(jsonMap); err != nil {
+		log.WithFields(log.Fields{
+			"body": string(body),
+		}).Debug(err.Error())
+
+		handlers.WriteErrorResponse(w, "Json did not match schema: "+path, http.StatusInternalServerError)
```
mogronalol: Badrequest error (400)

PR #318, opened 2016-12-08
Moved some middleware stuff onto a new Middleware struct, which is now used to store the middleware

benjih:
No functional changes, just preparing for changes.


danielbryantuk:
LGTM!

PR #337, opened 2016-12-24
GetAllEntries in memory cache not returning 'dest' as intented

shyal:
Hi Benji,

Not sure whether you should merge this in or not. But definitely worth investigating. Why are we creating this ``dest`` dict, yet not returning it? Moreover one has to wonder what it actually does? Is this a way create a copy? Sorry my level at go is still non-existent at this point.

Feel free not to merge this in. I'm really just creating this pull request to bring this to your attention.

Regards,

Shyal

### DIFF
--- a/core/cache/memory_cache.go
+++ b/core/cache/memory_cache.go
@@ -55,7 +55,7 @@ func (c *InMemoryCache) GetAllEntries() (map[string][]byte, error) {
 		dest[k] = v
 	}
 	c.RUnlock()
-	return c.elements, nil
+	return dest, nil
 }
 
 func (c *InMemoryCache) RecordsCount() (count int, err error) {


benjih:
Hey @shyal ,

Thanks for the pull request!

I had a look into this, you are correct. Its creating a copy to return to Hoverfly to use. 

Considering it wasn't using the copy and was working fine, the copying might not be needed...

shyal:
Again I'm not a go programmer, so I may be off here. Maybe you can help me through my thought process, and tell me if I'm wrong (I'm mostly thinking as a C / C++ programmer here).

```go
func (c *InMemoryCache) GetAllEntries() (map[string][]byte, error) {
	c.RLock()
	dest := make(map[string][]byte)
	for k, v := range c.elements {
		dest[k] = v
	}
	c.RUnlock()
	return c.elements, nil
}
```

It seems the copy is there because the `InMemoryCache` pointer `c`'s RW mutex needs to be locked for reading, which is what c.RLock() is doing.

```go
c.RLock()
```
At this point, no other thread can write to c (or more importantly to its elements). We then construct the dest variable using make built-in, which initialises and allocates it. Get to think of it this line:

```go
dest := make(map[string][]byte)
```
could be moved above `c.RLock()` for a small gain in efficiency (but who cares about speed, right? :)).

We then copy all the values into the new dictionary.  We then unlock c's mutex, so that the threads that were paused while waiting on the mutex can resume.

```go
c.RUnlock()
```

We finally return dest (in our new version anyway).

Returning c.elements without copying it means the caller of `GetAllEntries()` would receive an array that could potentially still be written to by other threads. The only reason why this hasn't lead to crashes is because we're reading from the dict returned by GetAllEntries(), as opposed to writing to it. Were we to write to it while other threads were resizing the memory cache, well at this point we'd be writing to other parts of program memory, which would most likely lead to a segfault, i.e a crash at a customer site.

But again, I don't know go.. so there may be some magic I'm unaware of. :)

shyal:
@benjih any thoughts?

benjih:
Yeah, you are right @shyal. I will merge this in soon.

PR #342, opened 2017-01-06
Improved logging around captureRequest and return error to avoid panics

benjih:


core/hoverfly.go
```diff
@@ -241,8 +241,15 @@ func (hf *Hoverfly) captureRequest(req *http.Request) (*http.Response, error) {
...
-			"error": err.Error(),
-			"mode":  "capture",
+			"error":       err.Error(),
+			"mode":        "capture",
+			"Path":        req.URL.Path,
+			"Method":      req.Method,
+			"Destination": req.Host,
+			"Scheme":      req.URL.Scheme,
+			"Query":       req.URL.RawQuery,
+			"Body":        string(reqBody),
```
rusenask: you will still get `panic` here. If an error occurs - probably just add something like:
`reqBody = []byte(") `
and log as much as possible (like you are doing now). Also, parameters from req.TLS struct could be useful as well for debugging

benjih: Done.

I had to log the TLS separately as it could be nil.

benjih:
Added the test and changed the new request to modifiedReq instead of req within the captureRequest function.

PR #348, opened 2017-01-10
Fix image and clarify export file

johnharris85:
Add some clarity to which file is created when exporting and fixed the image link on that page.


benjih:
Nice, thanks @johnharris85!

Issue #373, opened 2017-02-04
Can't use http_proxy or https_proxy directly with hoverfly


Bruno-DaSilva:
Currently we are running into a situation where we are blocked with a firewall to the service we want to mock, but we are able to access it via proxy server. Our normal setup to access this server is: 

app_server ------> proxy_server ------> service_to_mock

Now, in order to mock this, we require this setup:

app server ------> hoverfly ------> proxy_server ------> service_to_mock

However, hoverfly does not automatically configure this additional proxy server when http_proxy or https_proxy environment variables are set. Would it be possible to add in this functionality?

Bruno-DaSilva:
I've experimented with a small change on my end, and it seems to work perfectly. I will open up a PR, please let me know what you think of this change.

benjih:
Hey @Bruno-DaSilva, thanks for raising the issue and submitting a pull request. I will look into your pull request today.

Bruno-DaSilva:
PR merged, I consider this issue closed. Thanks @benjih.

PR #374, opened 2017-02-04
Added ProxyFromEnvironment configuration option 

Bruno-DaSilva:
This should allow auto proxy configuration based off of http_proxy/https_proxy and no_proxy environment variables, referenced in issue #373

### DIFF
--- a/core/cmd/hoverfly/main.go
+++ b/core/cmd/hoverfly/main.go
@@ -83,7 +83,7 @@ var (
 
 	tlsVerification = flag.Bool("tls-verification", true, "turn on/off tls verification for outgoing requests (will not try to verify certificates) - defaults to true")
 
-	externalProxy = flag.String("external-proxy", "", "specify an external proxy for hoverfly to route traffic through")
+	upstreamProxy = flag.String("upstream-proxy", "", "specify an upstream proxy for hoverfly to route traffic through")
 
 	databasePath = flag.String("db-path", "", "database location - supply it to provide specific database location (will be created there if it doesn't exist)")
 	database     = flag.String("db", inmemoryBackend, "Persistance storage to use - 'boltdb' or 'memory' which will not write anything to disk")
@@ -208,8 +208,8 @@ func main() {
 	}
 
 	// overriding environment variable (external proxy)
-	if *externalProxy != "" {
-		cfg.ExternalProxy = *externalProxy
+	if *upstreamProxy != "" {
+		cfg.UpstreamProxy = *upstreamProxy
 	}
 
 	// development settings
--- a/core/cmd/hoverfly/main.go
+++ b/core/cmd/hoverfly/main.go
@@ -83,6 +83,8 @@ var (
 
 	tlsVerification = flag.Bool("tls-verification", true, "turn on/off tls verification for outgoing requests (will not try to verify certificates) - defaults to true")
 
+	externalProxy = flag.String("external-proxy", "", "specify an external proxy for hoverfly to route traffic through")
+
 	databasePath = flag.String("db-path", "", "database location - supply it to provide specific database location (will be created there if it doesn't exist)")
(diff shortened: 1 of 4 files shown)


benjih:
Hey @Bruno-DaSilva,

The common usage of Hoverfly is set the HTTP_PROXY and HTTPS_PROXY environment variables to the address of Hoverfly.  This is something suggested to users in our documentation as well as being something our functional tests do.

As a result of this, this breaks functional tests by causing a proxying loop when the HTTP_PROXY and HTTPS_PROXY variables are set to Hoverfly.

There are two possible solutions to this.

The first is to implement some kind of check to make sure that those variables are not set to point to Hoverfly before setting the default HTTP client's proxy settings.

The second is to have proxy as a startup flag on the Hoverfly binary `core/hoverfly/cmd/main.go` which is then passed through GetNewHoverfly() and into GetDefaultHoverflyHTTPClient(). This startup flag could then be added to Hoverctl, providing the same functionality as this change, but more as a configuration.

I think the second option is more work, but it makes sense. If you are running behind a proxy but you wanted to put Hoverfly infront of that, you should be setting HTTP_PROXY and HTTPS_PROXY to Hoverfly and then telling Hoverfly that their is a proxy between itself and its intended destination.

I'm interested to hear your thoughts.

I'm interested to hear your opinion

Bruno-DaSilva:
Hey @benjih, you certainly bring up an excellent point. I've been looking at this from my own perspective only, where we use docker containers to isolate our services (and so there would not be conflicts with environment variables). You are right, though, that this does pose a problem for the larger user base and the solution I've offered thus far would potentially mean more work/confusion for them.

With regards to your first solution, while it could work, I do see a rather large problem with it. How does hoverfly know that the HTTP_PROXY env variable is pointing to itself? Does it send an http request to whatever it is set to in order to verify if it is actually pointing to a hoverfly instance? What happens if you're trying to chain multiple hoverflys together (it's a rather edge case circumstance but it _could_ happen), how would you know that the hoverfly you get a response back is yourself?

The second option, however, is much more explicit and less brittle in my opinion. You will avoid the 'autodetection' issue entirely, and give users full control of having a proxy setting enabled or disabled. There are two ways to do this in my opinion:
- have simply a true/false flag as an option which just takes from the env variables
- have an explicit setting somewhere (command line/config file; proxy_host: test.host.proxy.com, proxy_port: 443). It would require more implementation but would allow users to be explicit with their proxy settings without having to set or change an environment variable. 

Now, I think the second option is without a doubt the best solution of the two. My only question is which way to go (true/false vs explicit config) do you think is the better solution? I can take a crack at whichever you think is best.

benjih:
I think having it as an explicit option would be best and then have main.go pass it through on construction of a new Hoverfly instance via the config. Putting the proxy address in config seems appropriate to me.

Bruno-DaSilva:
Oh perfect, implementing that in hoverctl would have been my next question. I'll get on that :)

Bruno-DaSilva:
Hey @benjih, I've made the changes and updated the pull request (it's a super small change).

I'm new to golang so I'm sure there's some conventions I'm breaking. I'm open to any feedback as to what is missing/should be changed.

benjih:
Just had a looking at the pull request and its looking great!

I spoke to a few of the people here and we had a bit of a debate as to whether or not the `external-proxy` flag was clear enough. Someone mentioned that it might cause some confusion as external could have several meanings depending on the situation.

An idea was suggested to rename it to `upstream-proxy` as this was a bit cleare. If you configure a second proxy in a chain of proxies, you can only link them by configuring a proxy with an upstream proxy location.

I'm interested to hear your thoughts in the matter.

Other than that, all good. I await your response and any changes required and then I will merge this in.

Thank you very much @Bruno-DaSilva.

Bruno-DaSilva:
@benjih Thanks for taking a look so fast. You're definitely right that it's not clear enough, I was struggling to find a succinct flag to describe what it actually did. `upstream-proxy` sounds much better, I'll make that change.

PR #391, opened 2017-02-10
Added upstream-proxy flag to hoverctl and its config

benjih:


hoverctl/wrapper/config.go
```diff
@@ -177,6 +185,10 @@ func (this Config) BuildFlags() Flags {
 		flags = append(flags, "-tls-verification=false")
 	}
 
+	if this.HoverflyUpstreamProxy != "" {
```
tommysitu: Is this the same as len(this.HoverflyUpstreamProxy) >0

benjih: Yeah, these are the two common ways that you'd check a string to be empty. Both styles appear throughout the standard Golang packages.

We should probably stick to one of the two styles.

PR #429, opened 2017-02-27
Remove v1 records api

benjih:


benjih:
A lot of this is also updating the functional tests.

When doing assertions, it would check which is is persisted in Hoverfly, but it was using the v1 records endpoint.

core/import.go
```diff
@@ -92,24 +92,19 @@ func (hf *Hoverfly) ImportFromDisk(path string) error {
 		return fmt.Errorf("Got error while opening payloads file, error %s", err.Error())
 	}
 
-	var requests v1.RequestResponsePairPayload
+	var simulation v2.SimulationView
```
tommysitu: Does it mean importing v1 schema is not supported any more?

benjih: Yes, its been deprecated for months.

tommysitu: If people wants to migrate from v1 to v2, what's the option?

benjih: Before updating Hoverfly and hoverctl, import their simulation with `hoverctl import simulation.json --v1` and then `hoverctl export simulation.json`.

Hoverfly has been outputting v2 simulations by default since v0.9.0, so this would only affect people that created simulations before November of last year that haven't updated their simulation.

benjih: These instructions will also be posted with the release.

PR #430, opened 2017-02-27
Removed request type and treat everything as a template

benjih:


core/import.go
```diff
@@ -145,58 +145,58 @@ func (hf *Hoverfly) ImportRequestResponsePairViews(pairViews []interfaces.Reques
...
-			} else {
-				failed++
-			}
+			hf.Simulation.Templates = append(hf.Simulation.Templates, requestTemplateResponsePair)
+			success++
+			continue
+			// }
+
+			// if len(pair.Request.Headers) == 0 {
+			// 	pair.Request.Headers = make(map[string][]string)
```
tommysitu: Should they be deleted?

PR #431, opened 2017-02-27
Bypass RequestMatcher and use TemplateMatcher directly

benjih:
This is the first part of the work of changing the key value store from being a datastore for request recordings to becoming a cache for templates.

Template matching moves out of the RequestMatcher as this will in the future transform into wrapper around the cache when it is reimplemented.

### DIFF
--- a/core/hoverfly.go
+++ b/core/hoverfly.go
@@ -59,7 +59,6 @@ func GetNewHoverfly(cfg *Configuration, requestCache, metadataCache cache.Cache,
 	requestMatcher := matching.RequestMatcher{
 		RequestCache: requestCache,
 		Webserver:    &cfg.Webserver,
-		Simulation:   simulation,
 	}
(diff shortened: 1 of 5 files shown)


core/hoverfly.go
```diff
@@ -215,7 +214,22 @@ func (hf *Hoverfly) DoRequest(request *http.Request) (*http.Response, error) {
...
+		log.WithFields(log.Fields{
+			"error":       err.Error(),
+			"query":       requestDetails.Query,
+			"path":        requestDetails.Path,
+			"destination": requestDetails.Destination,
+			"method":      requestDetails.Method,
+		}).Warn("Failed to find matching request template from template store")
+
+		return nil, &matching.MatchingError{
+			StatusCode:  412,
```
tommysitu: Should it be 502?

PR #441, opened 2017-03-09
Switched out the strings in request templates for RequestFieldMatchers

benjih:


core/hoverfly.go
```diff
@@ -296,13 +296,25 @@ func (hf *Hoverfly) Save(request *models.RequestDetails, response *models.Respon
...
 	pair := models.RequestTemplateResponsePair{
 		RequestTemplate: models.RequestTemplate{
-			Path:        util.StringToPointer(request.Path),
-			Method:      util.StringToPointer(request.Method),
-			Destination: util.StringToPointer(request.Destination),
-			Scheme:      util.StringToPointer(request.Scheme),
-			Query:       util.StringToPointer(request.Query),
-			Body:        util.StringToPointer(request.Body),
-			Headers:     request.Headers,
+			Path: &models.RequestFieldMatchers{
```
tommysitu: Can you not keeping the existing method, but change its implementation to use the new method that take matchers? So that you don't need to modify all the existing code.

PR #448, opened 2017-03-13
Added a regexmatch for requests

benjih:


functional-tests/testdata.go
```diff
@@ -153,6 +153,34 @@ var JsonMatchSimulation = `{
     }
 }`
 
+var RegexMatchSimulation = `{
```
mogronalol: Maybe an example of regex where there is no match?

PR #455, opened 2017-03-14
Matching will now happen on all matchers, not the first one

benjih:


core/matching/field_matcher.go
```diff
@@ -1,72 +1,41 @@
...
-
-	"github.com/ChrisTrenkamp/goxpath"
-	"github.com/ChrisTrenkamp/goxpath/tree/xmltree"
-	"github.com/NodePrime/jsonpath"
 	"github.com/SpectoLabs/hoverfly/core/models"
-	glob "github.com/ryanuber/go-glob"
 )
 
 func FieldMatcher(field *models.RequestFieldMatchers, toMatch string) bool {
+	matches := []bool{}
```
mogronalol: You can get rid of this, and then just return `false` whenever there isn't a match, otherwise return `true`

Issue #460, opened 2017-03-15
Remote middleware requests are missing a Content-Type header


Bruno-DaSilva:
Hi @benjih,

Looks like hoverfly isn't sending a Content-Type header along with requests. I'm using a Spring Boot app as my middleware, but that defaults no Content-Type to application/octet-stream (with no way to configure otherwise as far as I can tell).

It's a quick code fix to add this, could you review the PR I've opened, please?

benjih:
Hey @Bruno-DaSilva, this has been merged. It will be released with v0.11.0 which is looking to be next week.

benjih:
[v0.11.0](https://github.com/SpectoLabs/hoverfly/releases/tag/v0.11.0) has now been released.

PR #461, opened 2017-03-15
Added application/json header to outgoing remote middleware requests

Bruno-DaSilva:
Unfortunately, Spring Boot defaults a missing Content-Type header to application/octet-stream. This should add an outgoing Content-Type: application/json header to every middleware requests.


Bruno-DaSilva:
Issue: #460

PR #465, opened 2017-03-16
Capture will now not record request headers by default

benjih:
If you want to capture headers, you will now need to provide an argument when setting the mode.

*For all request headers, provide a headers argument with the value of an asterik*
```
{
    "mode": "capture",
    "arguments: {
        "headers": "*"
    }
}
```

*For specific request headers, provide the header name (comma separated list)*
```
{
    "mode": "capture",
    "arguments: {
        "headers": "Content-Type"
    }
}
```

### DIFF
--- a/core/hoverfly.go
+++ b/core/hoverfly.go
@@ -292,7 +292,7 @@ func (hf *Hoverfly) GetResponse(requestDetails models.RequestDetails) (*models.R
 }
 
 // save gets request fingerprint, extracts request body, status code and headers, then saves it to cache
-func (hf *Hoverfly) Save(request *models.RequestDetails, response *models.ResponseDetails) error {
+func (hf *Hoverfly) Save(request *models.RequestDetails, response *models.ResponseDetails, headersToSave []string) error {
 	body := &models.RequestFieldMatchers{
 		ExactMatch: util.StringToPointer(request.Body),
 	}
--- a/core/hoverfly.go
+++ b/core/hoverfly.go
@@ -292,7 +292,7 @@ func (hf *Hoverfly) GetResponse(requestDetails models.RequestDetails) (*models.R
 }
 
 // save gets request fingerprint, extracts request body, status code and headers, then saves it to cache
-func (hf *Hoverfly) Save(request *models.RequestDetails, response *models.ResponseDetails, headersToSave []string) error {
+func (hf *Hoverfly) Save(request *models.RequestDetails, response *models.ResponseDetails, headersWhitelist []string) error {
 	body := &models.RequestFieldMatchers{
 		ExactMatch: util.StringToPointer(request.Body),
 	}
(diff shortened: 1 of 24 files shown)


core/hoverfly.go
```diff
@@ -307,6 +307,20 @@ func (hf *Hoverfly) Save(request *models.RequestDetails, response *models.Respon
 		}
 	}
 
+	var headers map[string][]string
+
+	if headersToSave != nil && len(headersToSave) == 0 {
```
mogronalol: I don't actually follow this. It looks like you are saving all headers whether it's an empty slice or a `nil` slice. I would see:

- Nil list means there is no whitelist, so save everything
- Empty list means there is a whitelist, containing no entries, so save nothing
- List with elements means these headers are whitelisted, so save them only

core/hoverfly_test.go
```diff
@@ -554,6 +553,104 @@ func Test_Hoverfly_Save_SavesRequestAndResponseToSimulation(t *testing.T) {
...
+
+	unit := NewHoverflyWithConfiguration(&Configuration{})
+
+	unit.Save(&models.RequestDetails{
+		Headers: map[string][]string{"testheader": []string{"testvalue"}},
+	}, &models.ResponseDetails{
+		Body:    "testresponsebody",
+		Headers: map[string][]string{"testheader": []string{"testvalue"}},
+		Status:  200,
+	}, nil)
```
mogronalol: It feels like this should be the other way round with the use of `nil` and empty `slice`. See my previous comment.

core/hoverfly_test.go
```diff
@@ -554,6 +553,104 @@ func Test_Hoverfly_Save_SavesRequestAndResponseToSimulation(t *testing.T) {
...
+
+	Expect(unit.Simulation.Templates[0].RequestTemplate.Headers).To(BeEmpty())
+}
+
+func Test_Hoverfly_Save_SavesAllRequestHeadersWhenGivenEmptyHeadersArray(t *testing.T) {
```
mogronalol: Same as last comment, other way around.

core/hoverfly.go
benjih: Yeah, its a bit confusing, CaptureMode, which is calling this function is looking to see if the headers to save is '*' and then just creating an empty list.

I'll move that logic down into here so it makes this code more readable.

core/hoverfly_service.go
```diff
@@ -66,6 +67,10 @@ func (this *Hoverfly) SetModeWithArguments(modeView v2.ModeView) error {
 		this.CacheMatcher.FlushCache()
 	}
 
+	if len(modeView.Arguments.Headers) > 1 && modeView.Arguments.Headers[0] == "*" {
```
mogronalol: I think the second expression should be `contains` so it checks every element.

PR #466, opened 2017-03-16
Using hoverctl, you can now specify which headers you want to capture

benjih:
When using capture mode, capturing headers is now disabled by default. If you need to record headers you should either do
```
hoverctl mode capture --all-headers
```
or
```
hoverctl mode capture --headers "Content-Type,User-Agent"
```

### DIFF
--- a/hoverctl/cmd/mode.go
+++ b/hoverctl/cmd/mode.go
@@ -39,13 +39,10 @@ mode is shown.
 
 				headersMessage = "and will capture all request headers"
 			} else if len(specficHeaders) > 0 {
-				modeView.Arguments.Headers = append(modeView.Arguments.Headers, strings.Split(specficHeaders, ",")...)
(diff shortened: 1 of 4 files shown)


hoverctl/cmd/mode.go
```diff
@@ -24,14 +29,37 @@ mode is shown.
...
+
+			var headersMessage string
+			if allHeaders {
+				modeView.Arguments.Headers = append(modeView.Arguments.Headers, "*")
+
+				headersMessage = "and will capture all request headers"
+			} else if len(specficHeaders) > 0 {
+				modeView.Arguments.Headers = append(modeView.Arguments.Headers, strings.Split(specficHeaders, ",")...)
+
+				if strings.Contains(specficHeaders, ",") {
```
tommysitu: Maybe removing the if else statement, and just print one message with 'request header(s')?

PR #493, opened 2017-04-04
Updated hoverctl destination to use targets

benjih:


functional-tests/hoverctl/hoverctl_suite_test.go
```diff
@@ -47,6 +47,10 @@ var _ = BeforeSuite(func() {
 	Expect(err).ShouldNot(HaveOccurred())
 })
 
+var _ = BeforeEach(func() {
```
benjih: Yeah, probably. When all of this code was originally written, it followed the examples provided by Ginkgo. As a result, I've added this each in the same style as BeforeSuite() and AfterSuite() functions.

PR #496, opened 2017-04-06
hoverctl mode now uses targets

benjih:


hoverctl/wrapper/hoverfly.go
```diff
@@ -103,6 +103,20 @@ func DeleteSimulations(target TargetHoverfly) error {
 }
 
 // GetMode will go the state endpoint in Hoverfly, parse the JSON response and return the mode of Hoverfly
+func GetMode(target TargetHoverfly) (string, error) {
+	response, err := doRequest(target, "GET", v2ApiMode, "")
+	if err != nil {
```
tommysitu: If no error, return empty string?

PR #500, opened 2017-04-07
Updating hoverctl start, stop and logs to use targets

benjih:


hoverctl/wrapper/hoverfly.go
```diff
@@ -597,6 +460,18 @@ func doRequest(target Target, method, url, body string) (*http.Response, error)
 	return response, nil
 }
 
+func checkPorts(ports ...int) error {
+	for _, port := range ports {
+		server, err := net.Listen("tcp", ":"+strconv.Itoa(port))
+		if err != nil {
+			return fmt.Errorf("Could not start Hoverfly\n\nPort %v was not free", port)
```
tommysitu: Can you check if we don't' need to close server here?

PR #507, opened 2017-04-12
Updated hoverctl targets command

benjih:
`hoverctl targets create` is now `hoverctl targets new`

Both `hoverctl targets new` and `hoverctl targets delete` will now take an argument instead of a flag for targetName.

```
hoverctl targets new new-target --admin-port 1234
```

### DIFF
--- a/functional-tests/hoverctl/config_test.go
+++ b/functional-tests/hoverctl/config_test.go
@@ -17,13 +17,13 @@ var _ = Describe("When I use hoverctl", func() {
 		})
 
 		It("prints the contents of the config.yaml", func() {
-			functional_tests.Run(hoverctlBinary, "targets", "create")
+			functional_tests.Run(hoverctlBinary, "targets", "create", "config-test")
 
 			output := functional_tests.Run(hoverctlBinary, "config")
 
 			Expect(output).To(ContainSubstring(`targets:`))
-			Expect(output).To(ContainSubstring(`default:`))
-			Expect(output).To(ContainSubstring("name: default"))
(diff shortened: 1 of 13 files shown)


functional-tests/hoverctl/config_test.go
```diff
@@ -17,13 +17,13 @@ var _ = Describe("When I use hoverctl", func() {
 		})
 
 		It("prints the contents of the config.yaml", func() {
-			functional_tests.Run(hoverctlBinary, "targets", "create")
+			functional_tests.Run(hoverctlBinary, "targets", "new", "config-test")
```
tommysitu: Why change to `new`? Having a keyword that reflects an action like these is good: 
``` 
hoverctl targets create target-name
hoverctl targets update target-name
hoverctl targets delete target-name
 ```

benjih: I changed it to new to be similar to the flag --new-target of which it is an alternative to. Pretty sure this was part of the changes we spoke about and agreed to last week, having the targets stuff looking similar to the flags that can be used on `hoverctl start` and `hoverctl login`.

tommysitu: There were a bit discussions last week, I think at one point I mentioned using the create/update/delete subcommands.  TBH, I don't remember what the conclusion was. If this is to match the --new-target flag, that's fine.

benjih:
@tommysitu Reverted the renaming of `hoverctl targets create` to `hoverctl targets new`.

PR #525, opened 2017-04-27
Added a 10 minute timeout to authentication login when three failed attempts are made

benjih:
If a subsequent failed attempt is made, the timeout is reset for another 10 minutes. When a timeout occurs, the user will receive a `429 Too Many Requests`.

Timeout will also occur if someone tries to bruteforce the username and password via a proxy request instead of an API call, though this will return a 401 regardless how many attempts have been made.

Updated Hoverctl to include a nice error message when this error occurs using `hoverctl login`

Hoverfly UI doesn't need to be updated, when the error occurs, an error message saying `Authentication Error: 429 Too Many Requests`

### DIFF
--- a/core/authentication/auth_service_test.go
+++ b/core/authentication/auth_service_test.go
@@ -1,0 +1,65 @@
+package authentication_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/SpectoLabs/hoverfly/core/authentication"
+	. "github.com/onsi/gomega"
+)
+
+var timeLayout = "2006-01-02T15:04:05.000Z"
+
+func Test_HasReachFailedAttemptsLimit_ReturnsFalseIfAttemptsIsBelowOrEqualToLimit(t *testing.T) {
+	RegisterTestingT(t)
+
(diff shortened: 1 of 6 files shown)


core/authentication/auth_service_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	authentication.Attempts.Count = 3
+	Expect(authentication.HasReachedFailedAttemptsLimit(3, "10s")).To(BeFalse())
+}
+
+func Test_HasReachFailedAttemptsLimit_ReturnsTrueIfAttemptsIsAboveLimit(t *testing.T) {
+	RegisterTestingT(t)
+
+	authentication.Attempts.Count = 4
+	Expect(authentication.HasReachedFailedAttemptsLimit(3, "10s")).To(BeFalse())
```
mogronalol: Should this not end with `BeTrue()`

benjih: Ah yeah, that's a false positive. The reason it passes is because no date is set last failed, which would never happen in the real system.

I have updated the test.

PR #531, opened 2017-05-08
Added X-HOVERFLY-AUTHORIZATION header and -disable-standard-proxy-auth flag

benjih:
The disable-basic-auth flag has been removed due to it no longer being a requirement for Hoverfly Cloud.

Instead, we have introduced the new `X-HOVERFLY-AUTHORIZATION` header. The reason for this is it using a custom header ensures it is sent via the HTTPS tunnel when proxying requests to Hoverfly. Along with this change, we have introduced the `-disable-standard-proxy-auth` flag to Hoverfly to allow you to disable the standard `Proxy-Authorization` header, which can be sometimes sent outside of the HTTPS tunnel.

### DIFF
--- a/core/cmd/hoverfly/main.go
+++ b/core/cmd/hoverfly/main.go
@@ -75,7 +75,7 @@ var (
 	isAdmin         = flag.Bool("admin", true, "supply '-admin false' to make this non admin user (defaults to 'true') ")
 	authEnabled     = flag.Bool("auth", false, "enable authentication, currently it is disabled by default")
 
-	proxyAuthorizationHeader = flag.String("proxy-auth", "default", "Switch the Proxy-Authorization header from proxy-auth `Proxy-Authorization` to header-auth `X-HOVERFLY-AUTHORIZATION`")
+	proxyAuthorizationHeader = flag.String("proxy-auth", "proxy-auth", "Switch the Proxy-Authorization header from proxy-auth `Proxy-Authorization` to header-auth `X-HOVERFLY-AUTHORIZATION`. Switching to header-auth will auto enable -https-only")
 
 	generateCA = flag.Bool("generate-ca-cert", false, "generate CA certificate and private key for MITM")
 	certName   = flag.String("cert-name", "hoverfly.proxy", "cert name")
--- a/core/cmd/hoverfly/main.go
+++ b/core/cmd/hoverfly/main.go
@@ -68,13 +68,14 @@ var (
 	destination = flag.String("destination", ".", "destination URI to catch")
 	webserver   = flag.Bool("webserver", false, "start Hoverfly in webserver mode (simulate mode)")
 
-	addNew           = flag.Bool("add", false, "add new user '-add -username hfadmin -password hfpass'")
-	addUser          = flag.String("username", "", "username for new user")
-	addPassword      = flag.String("password", "", "password for new user")
-	addPasswordHash  = flag.String("password-hash", "", "password hash for new user instead of password")
-	isAdmin          = flag.Bool("admin", true, "supply '-admin false' to make this non admin user (defaults to 'true') ")
-	authEnabled      = flag.Bool("auth", false, "enable authentication, currently it is disabled by default")
-	disableProxyAuth = flag.Bool("disable-standard-proxy-auth", false, "disables the standard Proxy-Authorization header, only allowing `X-HOVERFLY-AUTHORIZATION`")
+	addNew          = flag.Bool("add", false, "add new user '-add -username hfadmin -password hfpass'")
+	addUser         = flag.String("username", "", "username for new user")
+	addPassword     = flag.String("password", "", "password for new user")
+	addPasswordHash = flag.String("password-hash", "", "password hash for new user instead of password")
+	isAdmin         = flag.Bool("admin", true, "supply '-admin false' to make this non admin user (defaults to 'true') ")
+	authEnabled     = flag.Bool("auth", false, "enable authentication, currently it is disabled by default")
+
+	proxyAuthorizationHeader = flag.String("proxy-auth", "default", "Switch the Proxy-Authorization header from proxy-auth `Proxy-Authorization` to header-auth `X-HOVERFLY-AUTHORIZATION`")
 
 	generateCA = flag.Bool("generate-ca-cert", false, "generate CA certificate and private key for MITM")
 	certName   = flag.String("cert-name", "hoverfly.proxy", "cert name")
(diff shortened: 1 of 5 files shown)


core/cmd/hoverfly/main.go
```diff
@@ -283,6 +282,8 @@ func main() {
 		requestCache = nil
 	}
 
+	cfg.DisableProxyAuthoriation = *disableProxyAuth
```
mogronalol: It's confusing to have this as a boolean. There are three options `proxyAuth` or `hoverflyHeader` or `disabled`. I think in retrospect that the default should be `disabled`. It's a lot simpler than multiple combinations of different options. For example, if you disable authentication it doesn't make sense to be able to configure whether we are using a header or not, as it's turned off.

benjih: The only options there are regarding authorization are `-auth` to turn on authentication and `-disable-standard-proxy-auth` to disable the standard Proxy-Authentication header. 

Trying to repurpose the `-auth` flag to be both a switch for authentication and which headers are used would be more confusing as it would mean that every time you want to turn on auth, you would have to specify which headers you want, as not providing a value would have the same meaning as not providing the flag.

```
hoverfly -auth -username benjih -password 123
```
The `-auth` flag would equal `""`

```
hoverfly
```
The `-auth` flag would also equal `""`

I think what you are proposing would be more confusing for standard use cases in which the user is operating in a secure network. Considering I think that this is a use case we only expect to occur for us when starting instances of Hoverfly in Hoverfly Cloud, I think what I have done is fine. 

The alternative would be forcing users to always start Hoverfly by specifying proxyAuth.
```
hoverfly -auth=proxyAuth -username benjih -password 123
```

benjih: I also see no harm in accepting either `Proxy-Authentication` or `X-Hoverfly-Authentication` when using default settings. Further encouraging me to implement this flag in this way.

mogronalol: I see what you mean. Why not just this?:

```
hoverctl start // no auth
hoverctl start auth -username foo -password bar // proxy auth
hoverctl start auth -username foo -password bar -type proxy-auth // proxy auth
hoverctl start auth -username foo -password bar -type header-auth // header auth
```

mogronalol: So it's just an extra argument to auth that defaults

benjih: That is hoverctl. This flag is not implemented in hoverctl. The reason for this that I wanted to quickly implement it intime for the Hoverfly Cloud private beta. 

I could change the `-disable-standard-proxy-auth` flag on Hoverfly to be something like `-proxy-auth-type`.

tommysitu: I think what @mogronalol suggested is quite clear. I don't see any point of having a `disable-standard-proxy-auth` flag. If hoverfly is started with header-auth enabled, then any attempt to use `Proxy-Authorization` header should fail.

mogronalol: I think there should be a one-to-one mapping between the `hoverctl` and `hoverfly` flags because it simplifies things from the users point of view. I actually remember having issues around this last week (the certificate flag being named differently across the ctl and binary). If you only want to implement it in Hoverfly for now then go for it.

Yes I think `-proxy-auth-type` with two string values is better than the boolean.

benjih: So the two options are `proxy-auth` and `header-auth`. Does `proxy-auth` accept both `Proxy-Authorization` and `X-HOVERFLY-AUTHORIZATION`? If not, this means that switching between local instances of Hoverfly and remote instances would become more of a pain as you would have to change the authorization header being sent each time.

mogronalol: It is binary decision. You are either using standard proxy authentication, which is http and https, or you are using the custom header authentication, which is https only.

benjih: But why not accept both authorization header? Is it not better to follow Postel's Law and accept both values and then disable the insecure option when configuring Hoverfly for a secure deployment?

I don't think having both would be confusing. We change the docs to use the `X-HOVERFLY-AUTHORIZATION` header but mention that by default, we still accept the `Proxy-Authorization` for those using Hoverfly locally and want to use curl.
```
curl --proxy user:pass@localhost:8888 http://hoverfly.io
```

core/cmd/hoverfly/main.go
```diff
@@ -68,13 +68,14 @@ var (
...
-	authEnabled      = flag.Bool("auth", false, "enable authentication, currently it is disabled by default")
-	disableProxyAuth = flag.Bool("disable-standard-proxy-auth", false, "disables the standard Proxy-Authorization header, only allowing `X-HOVERFLY-AUTHORIZATION`")
+	addNew          = flag.Bool("add", false, "add new user '-add -username hfadmin -password hfpass'")
+	addUser         = flag.String("username", "", "username for new user")
+	addPassword     = flag.String("password", "", "password for new user")
+	addPasswordHash = flag.String("password-hash", "", "password hash for new user instead of password")
+	isAdmin         = flag.Bool("admin", true, "supply '-admin false' to make this non admin user (defaults to 'true') ")
+	authEnabled     = flag.Bool("auth", false, "enable authentication, currently it is disabled by default")
+
+	proxyAuthorizationHeader = flag.String("proxy-auth", "default", "Switch the Proxy-Authorization header from proxy-auth `Proxy-Authorization` to header-auth `X-HOVERFLY-AUTHORIZATION`")
```
mogronalol: Could we do a choice of `proxy-authorization` or `hoverfly-header-authorization` with the default as `proxy-authorization`? I think `default` is a too generic a name.

mogronalol: Other than that I'd merge it

core/cmd/hoverfly/main.go
```diff
@@ -68,13 +68,14 @@ var (
...
-	authEnabled      = flag.Bool("auth", false, "enable authentication, currently it is disabled by default")
-	disableBasicAuth = flag.Bool("disable-basic-auth", false, "disable Basic authentication, requiring all authentication to use API tokens")
+	addNew          = flag.Bool("add", false, "add new user '-add -username hfadmin -password hfpass'")
+	addUser         = flag.String("username", "", "username for new user")
+	addPassword     = flag.String("password", "", "password for new user")
+	addPasswordHash = flag.String("password-hash", "", "password hash for new user instead of password")
+	isAdmin         = flag.Bool("admin", true, "supply '-admin false' to make this non admin user (defaults to 'true') ")
+	authEnabled     = flag.Bool("auth", false, "enable authentication, currently it is disabled by default")
+
+	proxyAuthorizationHeader = flag.String("proxy-auth", "proxy-auth", "Switch the Proxy-Authorization header from proxy-auth `Proxy-Authorization` to header-auth `X-HOVERFLY-AUTHORIZATION`. Switching to header-auth will auto enable -https-only")
```
mogronalol: Sorry, just noticed as well, in the description do you need to say what the possible options are?

benjih: It talks about both proxy-auth and header-auth being options

PR #534, opened 2017-05-09
Added text/plain support for Logs API

benjih:
Sending a Content-Type header with `text/plain` will result in plaintext logs

### DIFF
--- a/core/handlers/v2/logs_handler.go
+++ b/core/handlers/v2/logs_handler.go
@@ -1,6 +1,7 @@
 package v2
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"net/http"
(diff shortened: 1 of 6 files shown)


core/handlers/v2/logs_handler.go
```diff
@@ -35,15 +38,27 @@ func (this *LogsHandler) Get(w http.ResponseWriter, req *http.Request, next http
 
 	queryParams := req.URL.Query()
 	limitQuery, err := strconv.Atoi(queryParams.Get("limit"))
-	if err == nil {
-		logs = this.Hoverfly.GetFilteredLogsView(limitQuery)
+
```
tommysitu: I think the following (pseudocode) might be less verbose:
```
// query is an object will a set of filters, or empty if query param is not present
query = parse(queryParams) 
logs = this.Hoverfly.getLogs(query) 
logsView = logsViewMapper.toView(logs)
```
Obviously, I don't know how you implement this in Go, but this would be how I do it in Java.

PR #546, opened 2017-05-16
Updated hoverctl logs to use the logs API from Hoverfly

benjih:
This pull request will make hoverctl work with remote instances of Hoverfly as well as instances running in Docker.

Logs will be retrieved from the Logs API located at `/api/v2/logs`. It will now no longer write the logs to a file.


hoverctl/cmd/logs.go
```diff
@@ -19,21 +20,42 @@ Shows the Hoverfly logs.
...
+		}
+
+		for i := 0; i < len(logs); i++ {
+
+			if logs[i] != "" && logsPrinted[logs[i]] != "x" {
+				fmt.Println(logs[i])
+				logsPrinted[logs[i]] = "x"
+			}
+
+			if i == len(logs)-1 && followLogs {
```
mogronalol: Is this just doing a `GET` request every five seconds? Does it return a diff between the currently viewed logs and the new logs?

hoverctl/wrapper/hoverfly.go
```diff
@@ -188,8 +189,45 @@ func SetMiddleware(target Target, binary, script, remote string) (v2.MiddlewareV
...
+func GetLogs(target Target, format string) ([]string, error) {
+	headers := map[string]string{
+		"Content-Type": "text/plain",
+	}
+
+	if format == "json" {
+		headers["Content-Type"] = "application/json"
+	}
+
+	response, err := doRequest(target, "GET", v2ApiLogs, "", headers)
```
mogronalol: Does this API call just return all the logs?

hoverctl/wrapper/hoverfly.go
```diff
@@ -188,8 +189,45 @@ func SetMiddleware(target Target, binary, script, remote string) (v2.MiddlewareV
...
+	if err != nil {
+		return []string{}, err
+	}
+
+	defer response.Body.Close()
+
+	responseBody, _ := ioutil.ReadAll(response.Body)
```
mogronalol: Are you taking the JSON here and transforming it into a different format?

benjih: No, I am working with the JSON as a string and trimming off the bits I don't need to print.

The logs endpoint will return a JSON object that looks like
```
{
    "logs": [
         {
             "level": "info",
              "msg": "something"
         },
         {
             "level": "info",
              "msg": "something"
         }
    ]
}
```

Previously, `hoverctl logs` would print logs such as
```
{"level":"info","method":"GET","msg":"started handling request","remote":"[::1]:61802","request":"/api/v2/logs","time":"2017-05-16T11:26:22+01:00"}
```
So I am trimming `{"logs":[` from the start and `]}` from the end. I am also removing the `,` from the end of each log.

I could have unmarshalled  them into structs and then iterated through each of the individual log elements and marshalled them back into JSON. But I think what have done is simpler. I can change it to unmarshal and then marshal again.

hoverctl/cmd/logs.go
benjih: It doesn't return a diff. Each log printed will be stored as a map key so that a quick check can be made to ensure that we don't print duplicates.

hoverctl/wrapper/hoverfly.go
benjih: It will limit the logs based on a limit query flag. If the flag isn't set, a default is used.

hoverctl/cmd/logs.go
mogronalol: Okay - maybe not as this but as part of another PR, it could have a way of returning diffs. For example providing an `after` parameter, or an `id` of the last retrieved log. Maybe we could create an issue? Because otherwise, the longer it runs, the more network traffic there will be until this call becomes a bottleneck.

benjih: Yeah. Creating an issue around it might be better.

hoverctl/wrapper/hoverfly.go
mogronalol: Ideally, this would happen externally. So you can use the `Accept` header to say whether you want JSON or text back from the endpoint. `Accept: application/json` means it will return JSON `Accept: text/*` text. Then you can just move this logic to the endpoint

hoverctl/wrapper/hoverfly.go
mogronalol: What is the limit / how do you determine the value?

benjih: The limit which Hoverfly will impose is 500. This is currently set [https://github.com/SpectoLabs/hoverfly/blob/master/core/handlers/v2/logs_handler.go#L26](here)

mogronalol: So it always returns the last 500 logs, and then removes duplicates?

hoverctl/cmd/logs.go
mogronalol: Saying that in the cloud version we will have a problem where logs live in memory and keep growing forever. Do we roll them in any way?

hoverctl/wrapper/hoverfly.go
benjih: So the `/api/v2/logs` endpoint is doing that, but its using the `Content-Type` header rather than the `Accept`. I will raise a separate PR and fix that now.

When you ask for logs in a JSON format from `/api/v2/logs`, should they come back in a valid JSON object or should they be coming across coming back as plaintext but as JSON objects?. For example:
```
{
    "logs": [
        {
            "level":"info",
            "method":"GET",
            "msg":"started handling request",
            "remote":"[::1]:61802",
            "request":"/api/v2/logs",
            "time":"2017-05-16T11:26:22+01:00"
        },
        {
            "level":"info",
            "method":"GET",
            "msg":"started handling request",
            "remote":"[::1]:61802",
            "request":"/api/v2/logs",
            "time":"2017-05-16T11:26:22+01:00"
        }
    ]
}
```
```
{"level":"info","method":"GET","msg":"started handling request","remote":"[::1]:61802","request":"/api/v2/logs","time":"2017-05-16T11:26:22+01:00"}
{"level":"info","method":"GET","msg":"started handling request","remote":"[::1]:61802","request":"/api/v2/logs","time":"2017-05-16T11:26:22+01:00"}
```

hoverctl/cmd/logs.go
benjih: In Hoverfly, we don't roll the logs yet. Definitely outside of this PR.

hoverctl/wrapper/hoverfly.go
benjih: Hoverfly will return the last 500 logs. Hoverctl will then print each of these, store a copy. If `--follow`, call again, remove duplicates and print new logs.

hoverctl/wrapper/hoverfly.go
benjih: I have updated this now so if the format is JSON, it will unmarshal the JSON and then the individual JSON logs, avoiding any potentially dodgey string manipulation. If the format isn't JSON, it'll assume its plaintext, which is the default in Hoverctl and still split that by line.

Issue #516, opened 2017-04-18
API endpoint for killing Hoverfly


mogronalol:
Right now we are using PID's to kill a running Hoverfly process. It's going to be a lot less brittle if it can be done via an API endpoint, using the URL of the target Hoverfly stored within the config. We are also a lot less likely to run across PID related problems which change across platforms.

mogronalol:
@johnharris85 I've assigned this to you if you were looking for some guidance. If you don't have the time the no worries at all and we will implement it ourselves. This should be the start of resolving your container issues as it will no longer need to check for PID's.

johnharris85:
Sounds good, will take a look next week.

PR #553, opened 2017-05-23
Added an API endpoint for stopping Hoverfly located at /api/v2/shutdown

benjih:
https://github.com/SpectoLabs/hoverfly/issues/516

### DIFF
--- a/core/handlers/v2/shutdown_handler.go
+++ b/core/handlers/v2/shutdown_handler.go
@@ -3,7 +3,6 @@ package v2
 import (
 	"net/http"
 	"os"
-	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/SpectoLabs/hoverfly/core/handlers"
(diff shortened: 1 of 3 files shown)


core/handlers/v2/shutdown_handler.go
```diff
@@ -0,0 +1,32 @@
...
+		negroni.HandlerFunc(am.RequireTokenAuthentication),
+		negroni.HandlerFunc(this.Get),
+	))
+}
+
+func (this *ShutdownHandler) Get(w http.ResponseWriter, req *http.Request, next http.HandlerFunc) {
+	handlers.WriteResponse(w, []byte(""))
+	go func() {
+		log.Warning("Hoverfly will shut down in 10 seconds")
+		time.Sleep(time.Second * 10)
```
mogronalol: Why the sleep? Doesn't `hoverctl stop` need to call this, in which case it should be instant?

This shouldn't be a get request it should be a `DELETE` I think

benjih: Yeah, fair enough. I agree with both of these suggestions.

PR #560, opened 2017-05-26
Switched hoverctl stop to use new shutdown API

benjih:
No more pids.


hoverctl/wrapper/hoverfly.go
```diff
@@ -246,20 +249,30 @@ func Start(target *configuration.Target, hoverflyDirectory configuration.Hoverfl
...
 
-	target.Pid = cmd.Process.Pid
-
 	return nil
 }
 
-func Stop(target *configuration.Target, hoverflyDirectory configuration.HoverflyDirectory) error {
-	hoverflyProcess := os.Process{Pid: target.Pid}
-	err := hoverflyProcess.Kill()
+func Shutdown(target configuration.Target) error {
```
tommysitu: Maybe this method should be called `Stop`

PR #571, opened 2017-06-02
Strongest match

mogronalol:


core/matching/header_matcher.go
```diff
@@ -4,23 +4,27 @@ import (
...
 		}
 
 		toMatchHeaderValues, toMatchHeaderValuesFound := toMatch[strings.ToLower(matcherHeaderKey)]
 		if !toMatchHeaderValuesFound {
-			return false
+			return FieldMatchWithNoScore(false)
 		}
 
 		for _, matcherHeaderValue := range matcherHeaderValues {
+			fmt.Println(toMatchHeaderValues)
```
benjih: Intentional logging? If so, this should use the Logrus logger.

core/modes/simulate_mode.go
```diff
@@ -5,25 +5,43 @@ import (
...
+	}
+}
 
 func (this SimulateMode) Process(request *http.Request, details models.RequestDetails) (*http.Response, error) {
 	pair := models.RequestResponsePair{
 		Request: details,
 	}
 
-	response, matchingErr := this.Hoverfly.GetResponse(details)
+	response, matchingErr := this.Hoverfly.GetResponse(details, strings.ToUpper(this.MatchingStrategy) != "FIRST")
```
benjih: I'm not sure I like this. I think that maybe the decision as to which matching strategy is used should be made from within GetResponse, rather than forcing each caller of the `GetResponse()` function to have to determine it themselves. This is also a public function on the Hoverfly so it could potentially be in use by others.

Moving the decision into GetResponse() should also keep things simpler if we decide to adopt alternative matching strategies.

PR #582, opened 2017-06-07
Fixed bug where two empty strings would result in a false match in the JsonMatcher

benjih:
Originally, you could match a `jsonMatch: ""` against an empty request body. This was broken in v0.11.5.

This bug was introduced when updating the JsonMatcher to ignore the order of object fields.

### DIFF
--- a/core/matching/json_match.go
+++ b/core/matching/json_match.go
@@ -6,7 +6,7 @@ import (
 )
 
 func JsonMatch(matchingString string, toMatch string) bool {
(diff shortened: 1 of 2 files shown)


core/matching/json_match.go
```diff
@@ -6,6 +6,9 @@ import (
 )
 
 func JsonMatch(matchingString string, toMatch string) bool {
+	if matchingString == "" && toMatch == "" {
```
tommysitu: How about if matchingString == toMatch; return true;

PR #614, opened 2017-06-22
"timeStarted" field in journal is now formatted in time.RFC3393

benjih:


benjih:
Golang doesn't have a format in the time library for ISO 8601. They do look similar though there are slight differences.

This change to use RFC3393 was so that it matched the date time format being used in the simulation file.

tommysitu:
The format used by hoverfly `2017-06-22T13:18:08+01:00` is OK. It is quite common among javascript. To parse it in JAVA, use the built in DateTimeFormater which is `ISO_OFFSET_DATE_TIME`

PR #617, opened 2017-06-23
Added filtering to journal

benjih:
You can now filter the journal by sending a POST request to `/api/v2/journal`. 

```
{
  "request": {
    "destination": {
      "globMatch": "*hoverfly*"
    }
  }
}
```

Currently it will only accept the request field. Though we plan to later support filtering based on other fields available on journal entries.

### DIFF
--- a/functional-tests/core/ft_api_v2_journal_test.go
+++ b/functional-tests/core/ft_api_v2_journal_test.go
@@ -1,6 +1,7 @@
 package hoverfly_test
 
 import (
+	"bytes"
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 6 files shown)


functional-tests/core/ft_api_v2_journal_test.go
```diff
@@ -140,6 +141,75 @@ var _ = Describe("/api/v2/journal", func() {
...
+
+				responseJson, err := ioutil.ReadAll(res.Body)
+				Expect(err).To(BeNil())
+
+				var errorView handlers.ErrorView
+
+				err = json.Unmarshal(responseJson, &errorView)
+				Expect(err).To(BeNil())
+
+				Expect(errorView.Error).To(Equal("Malformed JSON"))
```
tommysitu: Maybe the error should be `empty request in search params`, because the request is still a valid JSON

PR #623, opened 2017-07-05
When importing a simulation with an unknown schema, report correct error

benjih:


core/handlers/v2/simulation_views.go
```diff
@@ -57,6 +58,8 @@ func NewSimulationViewFromResponseBody(responseBody []byte) (SimulationViewV2, e
 		}
 
 		simulationView = simulationViewV1.Upgrade()
+	} else {
```
mogronalol: Maybe suggest that they upgrade, or question if they have the latest version?

PR #643, opened 2017-07-31
Fix encoded path in capture mode

clubifaximatic:
## Why the fix
When working in capture mode, if the original path is encoded, for some encoded characters, the request made by hoverfly to the real server is using the uncoded path instead of the original encoded path.

I my case I found this problem with some queries when using a 3rd party library that was signing the request based on the path (ie. /user/name%3Abob). When using hoverfly in capture mode, the response sent to the server was uncoded (ie. /user/name:bob). Then the signature failed in the server cause the path was different.

Go introduced RawPath in version 1.5 to address this problem [URL struct](https://golang.org/src/net/url/url.go?s=9159:9635#L325)

## Test error
Yo can test locally using `hoverfly`, `nc` and `curl` in three different shells

1. start hoverfly (shell1)
```
hoverfly -capture -v
```
2. start a server in port 9999 (shell2)
```
nc -l 9999
```
3. send a request with the path encoded (shell3)
```
curl --proxy http://localhost:8500 http://localhost:9999/user/name%3Abob
```
4. Check the request made by hoverfly is wrong, using the path `/user/name:bob` instead of `/user/name%3Abob` (shell2)
```
GET /user/name:bob HTTP/1.1   <--- ERROR
Host: localhost:9999
User-Agent: curl/7.51.0
Accept: */*
Proxy-Connection: Keep-Alive
Accept-Encoding: gzip
```

## Test fix
For testing the patch use the same procedure but with the new compiled hoverfly
0. compile
```
make build
```
1. start hoverfly (shell1)
```
target/hoverfly -capture -v
```
2. start a server in port 9999 (shell2)
```
nc -l 9999
```
3. send a request with the path encoded (shell3)
```
curl --proxy http://localhost:8500 http://localhost:9999/user/name%3Abob
```
4. Check the request made by hoverfly is good, using the path `/user/name%3Abob` (shell2)
```
GET /user/name%3Abob HTTP/1.1   <--- GOOD
Host: localhost:9999
User-Agent: curl/7.51.0
Accept: */*
Proxy-Connection: Keep-Alive
Accept-Encoding: gzip
```

### DIFF
--- a/core/handlers/v2/simulation_views_v1.go
+++ b/core/handlers/v2/simulation_views_v1.go
@@ -187,6 +187,9 @@ type RequestDetailsView struct {
 //Gets Path - required for interfaces.RequestMatcher
 func (this RequestDetailsView) GetPath() *string { return this.Path }
 
+//Gets RawPath - required for interfaces.RequestDetailsView
+func (this RequestDetailsView) GetRawPath() *string { return nil }
+
 //Gets Method - required for interfaces.RequestMatcher
 func (this RequestDetailsView) GetMethod() *string { return this.Method }
 
--- a/core/handlers/v2/simulation_views_v1.go
+++ b/core/handlers/v2/simulation_views_v1.go
@@ -187,9 +187,6 @@ type RequestDetailsView struct {
 //Gets Path - required for interfaces.RequestMatcher
 func (this RequestDetailsView) GetPath() *string { return this.Path }
 
-//Gets RawPath - required for interfaces.RequestDetailsView
-func (this RequestDetailsView) GetRawPath() *string { return nil }
-
 //Gets Method - required for interfaces.RequestMatcher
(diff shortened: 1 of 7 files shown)


benjih:
Hey @clubifaximatic,

Thanks for the thorough pull request.

Is there any way we can fix this without adding a new field to the RequestDetailsView struct? Maybe the path value should always be the RawPath? Having duplicate path values on the same struct might add confusion.

clubifaximatic:
Hi @benjih,

I changed the PR to add only the minimum necessary changes for this to work and do not modify any struct.

Following your advice, it know keep RawPath as a Path and later this value is used for creating the NewRequest. Using it in the constructor URLs generates both, the good Path and RawPath in the newRequest

benjih:
Perfect! Thank you very much @clubifaximatic!

PR #654, opened 2017-09-21
Update middleware.go

ooktay:
When script has less than 5 lines I get this:
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/SpectoLabs/hoverfly/hoverctl/cmd.glob..func13(0xc98de0, 0xc4201415c0, 0x0, 0x4)
	/home/ubuntu/.go_workspace/src/github.com/SpectoLabs/hoverfly/hoverctl/cmd/middleware.go:67 +0x5c9
...


tommysitu:
Thank you for fixing the bug!

Issue #675, opened 2017-11-15
Update the Dockerfile so end users can specify startup mode


sherl0cks:
Currently, you can only spin the container up in proxy mode with https://github.com/SpectoLabs/hoverfly/blob/master/Dockerfile#L11

tommysitu:
@sherl0cks The Dockerfile we use to build Hoverfly docker image is this one actually: https://github.com/SpectoLabs/hoverfly/blob/master/core/Dockerfile
I have updated it to allow passing commands, so that you can start hoverfly in webserver mode using docker: 
`docker run -d -p 8888:8888 -p 8500:8500 spectolabs/hoverfly:master -webserver`
Once this PR https://github.com/SpectoLabs/hoverfly/pull/676 is merged.

sherl0cks:
ahh thanks for the redirect @tommysitu . looks like that PR is merged, closing this issue.

psisodia:
Can anyone help me here, I have created hoverfly base image and added my git repo of API recordings to it. I can get image to build and run but I am unable to get hoverctl running. I need to use hoverctl in proxy mode for recording of APIs

PR #676, opened 2017-11-16
#675 run hoverfly docker image with command line flags

tommysitu:


benjih:
I'm pretty sure the Dockerfile in the root isn't used.

tommysitu:
I will remove that to avoid confusion. Thanks, @benjih

Issue #648, opened 2017-08-30
Feature Request: "pass-through simulate mode"


mackantestar:
Hi,

Use Case:

For example a web application providing customer specific data for logged in customers

**client <--api--> web app** <--api to simulate/pass through--> back end 

We only have control of the **bold** part above

There are cases when we want to simulate only certain services for a particular logged in customer. Could be that one of the services is slow, shaky, not complete, or we want to easily change the response.
Simulating one of the services can be a way of achieving e2e scenarios.

All other services though we want to "pass-through" to the real end-point. 

Instead of inside the web app create logic or point one of the services to the simulated, its much easier to just set a proxy for the whole web app. And the proxy decides if simulating or pass-through will happen. 

Related scenario

Lets say we simulate all services that a particular customer needs while logged in. We still want to be able to log in with another customer and for this customer recieve real data

What do you think about this functionality?

Best Regards
Marcus

JohnFDavenport:
I can see that this would be useful in some cases.

From my own experience, I have worked with some legacy systems which for various reasons have no good equivalent test system. When testing new systems that use them it becomes necessary to populate them with test data to simulate new functionality in the front-end or web-app in this example. The test data is distinguished by a fake customer-id, bank account number or flight number.

However, I don't personally see it as core functionality. It may fall into the category of a new business application for Hoverfly which would help customers with legacy systems.

How important is to your business @mackantestar ?

mackantestar:
Main idea for us with using hoverfly is actually to be able to serve testdata, as we cant add test-data directly in the back end. 

So we record data for real customer(-id), 
then create a fake customer-id (in the authentication system), 
then we change in the simulation.json so that req-resp pairs will match calls with the fake id instead of the real id
hoverfly will respond instead of the real system for the fake-id. And we also have complete control of the  responses

In practice we have created a test-customer in the back end..

Of course when we later test we could do a separate deploy with hoverfly set as proxy, and another deploy without the proxy

So mocked-tests for one deploy, test with real data in another deploy 

With the  "pass-through simulate mode" we can instead just have one deploy for all our tests. Its just a matter of if we log in with fake och real customer

mogronalol:
We already have an issue for this #389 which I think covers the functionality you are after.

mackantestar:
Aha yes it does. Didn't see 

Thank you both

Issue #389, opened 2017-02-09
Introduce spy or stub options to Hoverfly


mogronalol:
This comes from this issue in [Hoverfly-Java](https://github.com/SpectoLabs/hoverfly-java/issues/47).

With Mockito in Java, you have the ability to `Spy` on real objects. This means instead of having a pure mock, you have the real object. In this case, individual methods are overridden to become stubs, leaving the the non-stubbed methods to work as normal.

This seems like something Hoverfly could do, but for http. So something like:

`hoverctl simulate --stub` // The same as how it currently behaves
`hoverctl simulate` // The same as above because stub is default
`hoverctl simulate --spy` // When there isn't a match, call the real service

This is a valid use case, in some situations. The one given was running the external dependency locally or as part of your build, and only wanting to change the behaviour of specific endpoints.

tommysitu:
Spy mode is introduced in this PR: https://github.com/SpectoLabs/hoverfly/pull/680

tommysitu:
Thank you for your contribution @3sunnysideup, Hoverfly v0.15.0 has a new `spy` mode.

PR #680, opened 2017-12-13
Spy mode

3sunnysideup:
Add spy mode to hoverfly - similar to simulate mode but calls real server when cache miss


tommysitu:
Thank you for the PR! The request for a spy mode has come up for a few times, so this is really useful, eg.
https://github.com/SpectoLabs/hoverfly/issues/389
https://github.com/SpectoLabs/hoverfly/issues/648

tommysitu:
I've fixed some functional tests, other than that, looks good to me!

Issue #695, opened 2018-02-08
hoverctl doesn't connect to a custom host or a custom port


MatousJobanek:
If I want to get for example status of a hoverfly running on either a custom port or a custom host I execute:
`hoverctl status --admin-port 12345 --host 10.0.0.1`
but I get 
`Could not connect to Hoverfly at localhost:8888`

tommysitu:
Hi @MatousJobanek, `hoverctl` already supports multiple hoverfly instances, please check this out: http://hoverfly.readthedocs.io/en/latest/pages/tutorials/advanced/remotehoverfly/remotehoverfly.html

You just need to create a target and set it as default first: 
`hoverctl targets create --host 10.0.0.1 --admin-port 12345 --set-default`

Any `hoverctl` commands will point to that target.

MatousJobanek:
Ok, I missed that doc - I followed the hoverctl help when I was trying to get the status:
```
$ ./hoverctl status --help
If Hoverfly is running, this command will show an overview
of the instance of Hoverfly. This includes reporting the
mode and middleware set.

Usage:
  hoverctl status [flags]

Global Flags:
      --admin-port int   A port number for the Hoverfly API/GUI. Overrides the default Hoverfly admin port (8888)
  -f, --force            Bypass any confirmation when using hoverctl
      --host string      A host on which a Hoverfly instance is running. Overrides the default Hoverfly host (localhost)
      --proxy-port int   A port number for the Hoverfly proxy. Overrides the default Hoverfly proxy port (8500)
      --set-default      Sets the current target as the default target for hoverctl
  -t, --target string    A name for an instance of Hoverfly you are trying to communicate with. Overrides the default target (default)
  -v, --verbose          Verbose logging from hoverctl
```
so I thought that it should be possible to just call:

`hoverctl status --admin-port 12345 --host 10.0.0.1`

to override the default hoverfly settings - exactly as it is written in there.

tommysitu:
Aww, that's misleading... I don't think `--admin-port`,  `--proxy-port`, `--host` are global flags. But `--target` is. So limiting the scope fo those three flags will prevent any misuse.

tommysitu:
To be precise, these flags are global to `hoverctl start`, `hoverctl targets` and `hoverctl login`, but not for any other commands.

PR #696, opened 2018-02-08
fix(#695): when hoverctl is initialized the custom parameters provided on command line are set

MatousJobanek:
setting custom `admin-port`, `proxy-port` & `host` to be able to execute:
`hoverctl status --admin-port 12345 --host 10.0.0.1`


MatousJobanek:
I'm closing this PR as the ports and host are not global flags https://github.com/SpectoLabs/hoverfly/issues/695#issuecomment-364386765

Issue #699, opened 2018-02-20
New Diff mode for comparing response from real service with simulation config


MatousJobanek:
It would be good to have a mode for calling external service and comparing response with currently stored simulation to detect differences between our version and what is the current service response. 

Hoverfly would call the real service and would load matching simulation config. When the service responds then it would compare the response with the simulation (if it is present) and create diff report. Hoverfly then would return the real response to the client and store the diff report. From the test, it should be possible to get all diff reports stored from the last time with all differences.

The communication flow would look like:
```
 test                client       hoverfly              service
invoke client ----->  |              |                     |
  |              send request ---->  |                     |
  |                   |          resend request   ---->    |
  |                   |          and find simulation     process
  |                   |              |           <-----  response
  |                   |          receive response          |
  |                   |          and compare               |
  |                   |   <---resend real response---      |
get comparison results  ---------->  |                     |
  |   <-------------------------   return diff             |
fail or passs                        |                     |
```
WDYT? Is it something that would make sense in the sense of the Hovefly project?

I have already started working on it and I have some implementation for both hoverfly and hoverfly-java projects. I'll send my implementation proposal as a PR probably tomorrow.

tommysitu:
I think this is a fantastic idea! One problem with using a simulation is that the data could be out of sync with the real service over time, and using diff mode would reveal if the simulation is still faithful. 

I will try to review your PR this week.

danielbryantuk:
I think the idea sounds awesome -- funnily enough, in both of the Hoverfly workshops that @mogronalol and I ran over the last two weeks people asked about this!

mogronalol:
Just noticed this, and it is indeed a great idea!

PR #700, opened 2018-02-21
feat(#699): introduced new Diff mode

MatousJobanek:
Added new Diff mode for calling an external service and comparing response with currently stored simulation to detect differences between our version and what is the current service response.

#### Comparison
The comparison is performed only when some matching simulation is found.  

#### Response diff
If it is possible, then the response headers and bodies are unmarshaled and the resulting JSONs compared. if unmarshalling ends with some error, then deepEqual is used.

#### Diff reports
If there are found some differences, then the report is created inside of the hoverfly and stored in a map where the key is some basic definition (method, host, path, query) of the related request. These reports are stored and provided when requested on the endpoint `/api/v2/diff`

##### Diff report example

```
(0)The "header/X-Oauth-Scopes" parameter is not same - the expected value was [[admin:org, admin:org_hook, admin:public_key, admin:repo_hook, delete_repo, gist, notifications, repo, user]], but the actual one [undefined]
(1)The "header/X-Accepted-Oauth-Scopes" parameter is not same - the expected value was [[repo]], but the actual one [undefined]
(2)The "header/Date" parameter is not same - the expected value was [[Tue, 30 Jan 2018 09:47:02 GMT]], but the actual one [[Wed, 21 Feb 2018 18:44:58 GMT]]
(3)The "header/X-Github-Request-Id" parameter is not same - the expected value was [[98F0:0C4B:1908B3:37F815:5A703F15]], but the actual one [[F967:43FA:CF222D:1984AC5:5A8DBE29]]
(4)The "header/Status" parameter is not same - the expected value was [[404 Not Found]], but the actual one [[200 OK]]
(5)The "header/Transfer-Encoding" parameter is not same - the expected value was [[chunked]], but the actual one [undefined]
(6)The "body/followers" parameter is not same - the expected value was [%!s(float64=18)], but the actual one [%!s(float64=13)]
(7)The "body/company" parameter is not same - the expected value was [Red Hat Czec], but the actual one [Red Hat Czech]
(8)The "body/public_repos" parameter is not same - the expected value was [%!s(float64=56)], but the actual one [%!s(float64=57)]
(9)The "body/email" parameter is not same - the expected value was [matousjobanek@gmail.com], but the actual one [%!s(<nil>)]
```

### What is missing (feel free to add anything else you are missing in this PR)

- [x] tests
- [x] documentation
- [x] hoverctl commands

fixes: #699


tommysitu:
Hi @MatousJobanek, just got a chance to start looking into this PR. I got an error when change to diff mode:

```
$ ./hoverctl mode diff
diff is not a valid mode
```

tommysitu:
Hi @MatousJobanek, I am still learning Go, and so probably not the best person to comment on the code quality. I hope @benjih could add some inputs on that.

From the functional perspective, I have a few suggestions: 
1.  The diff report is lacking a timestamp. It would be nice to see when the diff messages are created so that one can inspect the report and easily correlate the differences with any API events such as a new release.

2. The diff message could be more readable and machine friendly if we store them in a JSON structure, such as:
```
"diff": [
  {
    "request": {
    "method": "GET",
    "host": "test.com",
    "path": "/",
    "query": ""
  },
  "diffs": [
    {
      "time": "2018-03-06T14:35:56Z", 
      "results": [
        {
           "field": "status",
           "expected": 200, 
           "actual": 400 
        },
        {
           "field": "header/Transfer-Encoding",        
           "expected": null      // null value indicates this header value was not present in original simulation
           "actual": "chunked" 
        }
      ]
    }
  ]
}
```

3. `hoverctl` commands could be simplified, for example: 
- `hoverctl diff` instead of `hoverctl diff-store get-all`
- `hoverctl diff delete` instead of `hoverctl diff-store delete-all`

This is all my comments. please let me know what you think.

benjih:
Hey @MatousJobanek,

Love the PR. Having something similar for this was always an idea I had that I never got around to. A couple of points though:

- Would you mind rebasing your work and resolving the conflict?
- It sounds like @tommysitu had issue running it, would you mind also looking into this?
- I also agree with @tommysitu about returning JSON. We have had issues in the past where we haven't returned machine-friendly data.
 
When I envisioned this over a year ago, I imagined this having the diff endpoint, and endpoint would be responsible for diffing the simulation. And it would check the whole simulation in one go. I feel that having to check response individually by sending exact request could be more painful than providing the ability to diff all responses at once.

Is there a reason as to why you went with the approach you did?

We would be happy to pick up this PR and finish off the work we feel is required?

MatousJobanek:
Hi guys,
thanks a lot for your reviews. All your points are totally valid, I just didn't have time to get to it this week. 
I have already started working on it and I should finish it early next week.

MatousJobanek:
Hi guys, 
sorry for the delay

@tommysitu 
> I got an error when change to diff mode ...

fixed in aa36054 - forgot to do that

> The diff report is lacking a timestamp

support for it added as part of acf9408 and incorporated into the hoverctl output here 3661cb5 

> The diff message could be more readable and machine friendly if we store them in a JSON structure

implemented in acf9408

the json format looks like:
```json
{
   "diff":[
      {
         "request":{
            "method":"GET",
            "host":"api.github.com:443",
            "path":"/users/matousjobanek",
            "query":""
         },
         "diffReports":[
            {
               "timestamp":"2018-03-15T15:44:36+01:00",
               "diffEntries":[
                  {
                     "field":"header/Transfer-Encoding",
                     "expected":"[chunked]",
                     "actual":"null"
                  },
                  {
                     "field":"header/Status",
                     "expected":"[404 Not Found]",
                     "actual":"[200 OK]"
                  }
               ]
            }
         ]
      }
   ]
}
```

> // null value indicates this header value was not present in original simulation

fixed in 94f9922 but I have no idea how to make the `null` non-string value

> hoverctl commands could be simplified

fixed in da4739e 

@benjih 
> Would you mind rebasing your work and resolving the conflict?

done

> Is there a reason as to why you went with the approach you did?

nothing in particular. it just looked like the simplest solution from the implementation point of view for me - I didn't know the code (and implementation details) of hoverfly-core and I'm also a beginner with Go language :-).
To be honest, I don't fully understand your idea. What would the endpoint do and how it would be implemented? What would be the difference comparing to my solution?

tommysitu:
@MatousJobanek Thank you for addressing all the comments! I will try it out later.

tommysitu:
Hi @MatousJobanek, when I try `./hoverctl diff get`, it only prints the last report, not all of them. Is it supposed to do that?

MatousJobanek:
Weird, it should return all of them. Let me check it where the problem is.

MatousJobanek:
@tommysitu That was my mistake when I was creating the output string - it's fixed 2d6aa8e 
sorry for that

tommysitu:
Hey @MatousJobanek, thanks for the quick fix. We should be to merge it in soon. Well done!

Issue #693, opened 2018-02-07
How to set ip for hoverfly?


karlpokus:
Hi!

How do I set hoverfly to listen on something other than loopback like `10.0.0.0/8`? The only config I can find concerns `hoverctl` and it's remotes.

tommysitu:
Hi @karlpokus, you can set hoverfly to listen on a single address only, for example `hoverfly -listen-on-host 10.0.0.1` , or listen to all interfaces: `hoverfly -listen-on-host 0.0.0.0`. I don't think it works with a CIDR block.

karlpokus:
@tommysitu Wonderful! Thank you. I'm gonna try this.

karlpokus:
@tommysitu 

$ hoverfly -listen-on-host 10.0.0.0/8

INFO[2018-02-07T13:44:24+01:00] Listen on specific interface                  host=10.0.0.0/8
INFO[2018-02-07T13:44:24+01:00] Using memory backend                         
INFO[2018-02-07T13:44:24+01:00] Proxy prepared...                             Destination=. Mode=simulate ProxyPort=8500
INFO[2018-02-07T13:44:24+01:00] current proxy configuration                   destination=. mode=simulate port=8500
FATA[2018-02-07T13:44:24+01:00] Failed to start proxy                         error=listen tcp: lookup 10.0.0.0/8: no such host

karlpokus:
`0.0.0.0` works though.

tommysitu:
You are not the first one asking the question. I have added the answer to the "Troubleshooting" section in the docs now.

karlpokus:
@tommysitu Yeah documenting this sounds good. btw: any chance of adding the flag to hoverctl in the future?

tommysitu:
Yes, inconsistent flags between hoverctl and hoverfly is an issue. If you want to do some Go, feel free to send us a PR.

karlpokus:
@tommysitu I'd be happy to contribute. Let me see if I can find a spot. Regardless - thanks for the swift reply. Really like working with hoverfly.

benjih:
The inclusion of the `listen-on-host` flag has been merged in via #705 and is now awaiting release.

karlpokus:
@benjih Excellent news! Thank you 😃

PR #705, opened 2018-03-08
adding listen-on-host flag to hoverctl

ifuyivara:
hoverctl is missing some flags. This PR adds the -listen-on-host flag so you can specify the host binding. 

see https://github.com/SpectoLabs/hoverfly/issues/693 for more info on why this is being added


benjih:
Fantastic, thank you very much @ifuyivara. We appreciate the effort and are very delighted to hear how much you enjoy working with Hoverfly!

karlpokus:
@ifuyivara Thanks!

PR #706, opened 2018-03-12
Improve errors from middleware

benjih:
Errors from middleware have always been a bit lacking in information. To combat this, I have introduced a MiddlewareError which should wrap all the relevant bits of information you would want to allow you debug your middleware. This new error should be logged during a failed middleware execution during an in-flight request through Hoverfly. This error is also output via the API when setting middleware, resulting in in appearing in hoverctl. 

Please find an example below of the new style errors.

```
Testing middleware against Hoverfly...
Could not set middleware, it may have failed the test

Middleware failed
Command: python /tmp/hoverfly/hoverfly_575209900
exit status 1

STDIN:
{"response":{"status":200,"body":"ok","encodedBody":false,"headers":{"test_header":["true"]}},"request":{"path":"/","method":"GET","destination":"www.test.com","scheme":"","query":"","body":"","headers":{"test_header":["true"]}}}

STDERR:
Traceback (most recent call last):
  File "/tmp/hoverfly/hoverfly_575209900", line 31, in <module>
    main()
  File "/tmp/hoverfly/hoverfly_575209900", line 23, in main
    t = payload_dict['data']['pairs'][0]['request']['headers']['Authorization'][0]
KeyError: 'data'

```


JohnFDavenport:
Thanks. Looks good to me. It should be a big assistant to users.

core/middleware/middleware.go
```diff
@@ -103,27 +96,11 @@ func (this *Middleware) SetBinary(binary string) error {
 		this.Binary = ""
 		return nil
 	}
-	testCommand := exec.Command(binary)
```
tommysitu: I believe this is the duplicated middleware checking you mentioned. If we remove it, does it mean using `hoverfly` command to set middleware would have no checking done?

benjih: There is still checking done, this was the preliminary checks.

A test execute happens with the new middleware here:

https://github.com/SpectoLabs/hoverfly/pull/706/files/faaa6d794102efcf405bd1493d9d2827a9287d24#diff-e5ffef0351b641e124a65ce90d9017b4R161

PR #707, opened 2018-03-12
Exact Match is now worth two points

benjih:


tommysitu:
We need to update the docs.

PR #712, opened 2018-03-16
Using k8s.io/client-go/util/jsonpath instead of github.com/NodePrime/…

benjih:
…jsonpath


core/matching/json_path_match.go
```diff
@@ -1,19 +1,47 @@
...
 
 func JsonPathMatch(matchingString string, toMatch string) bool {
-	paths, err := jsonpath.ParsePaths(matchingString)
+	matchingString = prepareJsonPathQuery(matchingString)
+
+	jsonPath := jsonpath.New("")
+
+	err := jsonPath.Parse(matchingString)
 	if err != nil {
 		return false
```
tommysitu: It'll be good to add some debug logs before each return statements in this method. It will help to understand why the match failed, whether is invalid jsonpath expression or cannot unmarshal the `toMatch` string

core/matching/json_path_match.go
```diff
@@ -1,19 +1,47 @@
...
+	"k8s.io/client-go/util/jsonpath"
+)
 
 func JsonPathMatch(matchingString string, toMatch string) bool {
-	paths, err := jsonpath.ParsePaths(matchingString)
+	matchingString = prepareJsonPathQuery(matchingString)
+
+	jsonPath := jsonpath.New("")
+
+	err := jsonPath.Parse(matchingString)
```
tommysitu: It'll be good to add some debug logs before each return statements in this method. It will help to understand why the match failed, whether is invalid jsonpath expression or cannot unmarshal the `toMatch` string

Issue #717, opened 2018-03-22
Minor spelling error


somejeff:
When receiving an error, Hoverfly returns a message containing a spelling error.

```
Hoverfly Error!

There was an error when forwarding the request to the intended desintation
```

PR #718, opened 2018-03-22
#717 minor spelling

somejeff:


tommysitu:
@somejeff, thanks for the fix, appreciated!

PR #721, opened 2018-04-13
Headers with matchers

benjih:


core/hoverfly_funcs.go
```diff
@@ -52,39 +50,29 @@ func (hf *Hoverfly) GetResponse(requestDetails models.RequestDetails) (*models.R
...
 
-		strongestMatch := strings.ToLower(mode.MatchingStrategy) == "strongest"
-
 		// Matching
-		if strongestMatch {
-			pair, err, cachable = matching.StrongestMatchRequestMatcher(requestDetails, hf.Cfg.Webserver, hf.Simulation, hf.state)
-		} else {
-			pair, err, cachable = matching.FirstMatchRequestMatcher(requestDetails, hf.Cfg.Webserver, hf.Simulation, hf.state)
-		}
+		result := matching.Match(mode.MatchingStrategy, requestDetails, hf.Cfg.Webserver, hf.Simulation, hf.state)
```
tommysitu: This is better

core/matching/header_matcher.go
```diff
@@ -1,55 +1,60 @@
...
 		}
+		fmt.Println(toMatch)
+		fmt.Println("matcher key: " + matcherHeaderKey)
+		matcherHeaderValueMatched := false
 
-		toMatchHeaderValues, toMatchHeaderValuesFound := toMatch[strings.ToLower(matcherHeaderKey)]
-		if !toMatchHeaderValuesFound {
-			return FieldMatchWithNoScore(false)
+		toMatchHeaderValues, found := toMatch[strings.ToLower(matcherHeaderKey)]
+		if !found {
```
tommysitu: If the header key is not found, do you still need to invoke the `ScoredFieldMatcher`?

PR #731, opened 2018-06-01
Switching queries to be keyed like headers

benjih:
Moving old queries to `depricatedQuery` field.


tommysitu:
typo in "depricatedQuery"

Issue #697, opened 2018-02-13
Content-Length isn't recorded or returned


Fishbowler:
My organisation has a disperate collection of APIs consumed in combinations from various front-ends. Hoverfly could simplify testing by simulating one or more of these to focus on the "thing under test".

I've already got a set of API tests in place, and I'm using them to power the recording and test the simulation.

One endpoint accepts and validates some post-data, and when the data is correct, returns a 200 and no body. The API test validates that the Content-Length header is 0.

When recording through Hoverfly using the `--all-headers` switch, the Content-Length header isn't captured, and when simulating, the test fails. I've manually edited the export to include the header, but I'm not seing any change.

```
				"response": {
					"status": 200,
					"body": "",
					"encodedBody": true,
					"headers": {
						"Cache-Control": [
							"private"
						],
						"Content-Encoding": [
							"gzip"
						],
						"Date": [
							"Tue, 13 Feb 2018 08:25:14 GMT"
						],
						"Hoverfly": [
							"Was-Here"
						],
						"Last-Modified": [
							"Tue, 13 Feb 2018 08:25:14 GMT"
						],
						"Link": [
							""
						],
						"Transfer-Encoding": [
							"chunked"
						],
						"Vary": [
							"Accept-Encoding"
						],
						"Content-Length": [
							"0"
						]
					},
					"templated": false
				}
```

Is this header specifically excluded? Or excluded when 0?
Any advice on how to make this more "faithful"?

I'm using hoverctl and hoverfly v0.15.0, downloaded binaries, on the commandline.

tommysitu:
Interesting problem. Hoverfly is default to not capture any Request headers, but it should capture all Response headers. The '--all-headers' flag is to tell hoverfly to capture all Request headers therefore it won't make any difference here. 

Could you confirm that the response indeed contains the Content-Length header? For example making a request with curl and -v flag without hoverfly as a proxy.

Fishbowler:
I'm in a particularly useful scenario in that I'm fueling my recordings through a set of API tests.
With hoverfly, the Content-Length is missing and the test fails.
Without hoverfly, the Content-Lenght is present and 0 and the test passes.

tommysitu:
Ok, in that case I will do some debuggings when I have my laptop around next week.

tommysitu:
Hi @Fishbowler, I just realized this issue is possibly a duplicate of this one: https://github.com/SpectoLabs/hoverfly/issues/422

[...]

Issue #727, opened 2018-05-19
HTTPS proxy returns incompatible headers with node.js client


nordicdyno:
HTTPs proxy in capture and simulation mode returns both `Content-Length` and `Transfer-Encoding: chunked` headers. It' breaks node client (throws `HPE_UNEXPECTED_CONTENT_LENGTH` error).

OS: MacOS
Hoverfly version:  v0.16.0
node: v8.9.0

JohnFDavenport:
Looks like Hoverfly shouldn't be setting the content length or your client should be ignoring it. We'll have a closer look.

tommysitu:
#https://github.com/SpectoLabs/hoverfly/issues/697

tommysitu:
It's related to this issue: https://github.com/SpectoLabs/hoverfly/issues/697

tommysitu:
I think meeting the following requirements from [RFC7230](https://tools.ietf.org/html/rfc7230) should solve the `Content-Length` and `Transfer-Encoding` issues:
1. "A sender MUST NOT send a Content-Length header field in any message that contains a Transfer-Encoding header field."
2. “In the absence of Transfer-Encoding, an origin server SHOULD send a Content-Length header field when the payload body size is known prior to sending the complete header section. “
3. "A server MUST NOT send a Transfer-Encoding/Content-Length header field in any
   response with a status code of 1xx (Informational) or 204 (No Content).
4. "A server MUST NOT send a Transfer-Encoding/Content-Length header field in
   any 2xx (Successful) response to a CONNECT request.”

The problem was that goproxy alwasy delete Content-Length header and set Transfer-Encoding header when handling HTTPS response for some reasons. 

Hoverfly on the other hand should have different behavior. In capture mode, it should just record the original headers faithfully. In simulation mode, I can think of two scenarios need special treatments: 
1. User does not provide Content-Length and Transfer-Encoding headers when the response has a body. This is quite common when the simulation is manually created, I think Hoverfly should add approriate header in this case. 
2. User modified the response body but not update the Content-Length header. This could happen when middlware was used to modify response body. Hoverfly should either work out the correct Content-Length or just set chucked transfer encoding.

benjih:
Hey @nordicdyno 

This should now be fixed in [Hoverfly v0.17.1](https://github.com/SpectoLabs/hoverfly/releases/tag/v0.17.1). If this is not the case and you still experience problems, please reopen this issue so that we can discuss further.

PR #743, opened 2018-06-26
Fix regression where Hoverfly returns Transfer-Encoding and Content-Length

benjih:
This regression was identified in https://github.com/SpectoLabs/hoverfly/issues/727 and came about trying to fix issue https://github.com/SpectoLabs/hoverfly/issues/697

### DIFF
--- a/functional-tests/core/ft_proxy_test.go
+++ b/functional-tests/core/ft_proxy_test.go
@@ -125,6 +125,44 @@ var _ = Describe("When I run Hoverfly", func() {
 			Expect(response.Header.Get("Content-length")).To(Equal("5555"))
 			Expect(response.Header.Get("Transfer-Encoding")).To(Equal(""))
 		})
+
+		It("should not set Content-Length if Transfer-Encoding set", func() {
+
+			hoverfly.ImportSimulation(`{
+				"data": {
+					"pairs": [
+						{
+							"request": {
+								"path": {
+									"exactMatch": "/path"
+								}
+							},
+							"response": {
+								"status": 200,
+								"body": "OK",
+								"headers": {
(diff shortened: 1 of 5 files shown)


functional-tests/core/ft_proxy_test.go
```diff
@@ -54,6 +55,76 @@ var _ = Describe("When I run Hoverfly", func() {
...
+
+			body, err := ioutil.ReadAll(response.Body)
+			Expect(err).To(BeNil())
+			Expect(string(body)).To(Equal("OK"))
+
+			Expect(response.Header.Get("Content-Length")).To(Equal("2"))
+			Expect(response.Header.Get("Transfer-Encoding")).To(Equal(""))
+		})
+
+		It("should not set Content-Length if not empty", func() {
```
tommysitu: Another useful test would be "should not set Content-Length if Transfer-Encoding header is present."

functional-tests/core/ft_proxy_test.go
```diff
@@ -54,6 +55,114 @@ var _ = Describe("When I run Hoverfly", func() {
...
+			Expect(response.StatusCode).To(Equal(http.StatusOK))
+
+			body, err := ioutil.ReadAll(response.Body)
+			Expect(err).To(BeNil())
+			Expect(string(body)).To(Equal("OK"))
+
+			// Should be empty as Transfer-Encoding was set
+			Expect(response.Header.Get("Content-length")).To(Equal(""))
+			// Will always be empty as they are excluded by net/http
+			Expect(response.Header.Get("Transfer-Encoding")).To(Equal(""))
```
tommysitu: We should find a way to validate this in the test somehow.

PR #745, opened 2018-06-27
Warn users when importing simulations with header problems

benjih:
Warn when:
 * Content-Length and Transfer-Encoding are set on a response
 * Content-Length does not match response body length``

I have also sneaked in some refactoring to errors. We have lots of different errors that kind of do the same thing, so I'm starting to unify errors.

### DIFF
--- a/core/errors/errors.go
+++ b/core/errors/errors.go
@@ -50,12 +50,6 @@ func ContentLengthMismatchError() *HoverflyError {
 	}
 }
 
-func ContentLengthAndTransferEncodingHeaderError() *HoverflyError {
-	return &HoverflyError{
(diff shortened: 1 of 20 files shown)


core/errors/errors.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (err HoverflyError) Error() string {
+	return err.Message
+}
+
+func NoCacheSetError() *HoverflyError {
+	return &HoverflyError{
+		Message:    "No cache set",
+		StatusCode: 412,
```
tommysitu: I don't think we should be using `Precondition Failed` status code, it's only applied to conditional requests with those `If-xxx` headers.

Issue #744, opened 2018-06-26
XPath matcher fails to parse XML documents missing XML declaration


drdivano:
XPath matcher fails to parse XML documents missing XML declaration (<?xml version="1.0" ?>).
Hovever XML declaration is optional in XML 1.0, and should not lead to parse failure

The following simulation reproduces the bug:

```
{
  "data": {
    "pairs": [
      {
        "request": {
          "method": [
            {
              "matcher": "exact",
              "value": "POST"
            }
          ],
          "body": [
            {
              "matcher": "xpath",
              "value": "/*"
            }
          ]
        },
        "response": {
          "body": "HELLO\n",
          "status": 200,
          "encodedBody": false,
          "templated": false
        }
      }
    ],
    "globalActions": {}
  },
  "meta": {
    "schemaVersion": "v5"
  }
}

```

This example works (returns HELLO):

```
https_proxy=https://10.100.51.173:8500 curl -k -X POST --data '<?xml?><q/>' https://example.com/xxx`
```

This one fails:

```
https_proxy=https://10.100.51.173:8500 curl -k -X POST --data '<q/>'  https://example.com/xxx

ERRO[2018-06-26T18:01:33Z] Failed to load XML tree: Malformed XML file
WARN[2018-06-26T18:01:33Z] Failed to find matching request from simulation  destination=example.com error=No match found method=POST path=/xxx query=map[]
```

JohnFDavenport:
We're just getting round to looking at this. Xpath is dependent on a third-party lib.

Issue #678, opened 2017-11-22
Passing a Parameter dynamically to response from request


priyathamreddy6009:
Hi Team,
We are trying to stub one Json response using hoverfly. Currently we are getting static response for any request. We are expecting to get one value dynamically(empID) in response that should be take from request.

Example:
Request: 
{"employee":{"empName":"Naren","empDesg":Test,"empID":"76443",}}

Response:
{
    "employee": {
        "empID":"76443"
    },
    "status": "success"
}
So in the above example the empID value keep changing for all employees. So we want to get the empID dynamic value in response while stubbing through Hoverfly. Can you please help us here.

priyathamreddy6009:
Can someone help me here?

JohnFDavenport:
You can use middleware to do that with Hoverfly in simulate mode. Both the request body and the corresponding response are available to the middleware. There is an example [here](https://docs.hoverfly.io/en/latest/pages/tutorials/basic/modifyingresponses/modifyingresponses.html#modifyingresponses) and there's a SOAP example in the hoverpy documentation [here](http://hoverpy.readthedocs.io/en/latest/pages/usage/modify_soap.html). @tommysitu was showing me a Java example but I can't find that at the moment. Hopefully he can find it and post it.

I think the documentation could be clearer. Some of the diagrams do not show middleware being available in simulate mode.

tommysitu:
@priyathamreddy6009 I need to point out that using request body for response templating is not supported yet, so the only way is using middleware as @JohnFDavenport suggested. You can have a look here to see what's available: https://hoverfly.readthedocs.io/en/latest/pages/keyconcepts/templating/templating.html

tommysitu:
@priyathamreddy6009 You should be able to use response templating to do this in the next release, your response body should look like this: 
```
"response": {
	"status": 200,
	"body": "{ \"employee\": { \"empID\": \"{{Request.Body \"jsonpath\" \"$.employee.empID\"}}\" }",
        "templated": true
}
```

Teklie:
@tommysitu I liked the response templating feature you mentioned above. Do you have a link to access its documentation please?

ns3777k:
@Teklie 
https://docs.hoverfly.io/en/latest/pages/keyconcepts/templating/templating.html

Teklie:
> @Teklie
> https://docs.hoverfly.io/en/latest/pages/keyconcepts/templating/templating.html

Thanks @ns3777k !

PR #755, opened 2018-07-16
Allow templating access to request body for use with jsonpath and xpath

benjih:
Also some functional tests and also request method as it was very easy. https://github.com/SpectoLabs/hoverfly/issues/678

This PR also changes XML parsing to be not strict, which should fix https://github.com/SpectoLabs/hoverfly/issues/744


core/templating/templating.go
```diff
@@ -47,6 +50,7 @@ func NewTemplator() *Templator {
 		raymond.RegisterHelper("randomIPv4", t.randomIPv4)
 		raymond.RegisterHelper("randomIPv6", t.randomIPv6)
 		raymond.RegisterHelper("randomUuid", t.randomUuid)
+		raymond.RegisterHelper("Request.body jsonPath", t.jsonPath)
```
tommysitu: Is it supposed to register {{ Request.Body xpath }} as well?

functional-tests/testdata/templating_helpers.go
```diff
@@ -0,0 +1,190 @@
+package testdata
+
+var TemplatingHelpers = `{
```
tommysitu: I don't think the test data is being used.

benjih: https://github.com/SpectoLabs/hoverfly/blob/templating-request-body/functional-tests/core/ft_templated_response_test.go#L120

core/templating/templating.go
benjih: This was an old attempt, its not used or required, should be deleted.

Issue #750, opened 2018-07-05
V5 schema does not support matching on an empty request query


tommysitu:
In the new V5 schema, the query field is a map of key and matchers. If you set it as an empty map, Hoverfly simply ignores the query for matching, and there is no way to create a simulation to match on an empty query. As a workaround, you have to use the `deprecatedQuery` field to create an exact matcher on an empty query string which is likely to be removed in the future.

PR #758, opened 2018-07-31
Support empty and nil queries

benjih:
During the implementation of the new v5 schema, support for matching specifically against a request with no query parameters was left off. This resulted in people who required this matching to continue to have to use the deprecated query field.

This PR should fix this issue.

Further documentation can be found in #750.


tommysitu:
LGTM

PR #759, opened 2018-07-31
Use headers mode arguments when setting diff mode

benjih:
It is now possible to set the headers to capture in diff mode, similarly to how it works in capture mode.

The default is now that it will not diff on headers. To use the diff mode as it was previously...
```
hoverctl mode diff --all-headers
```


core/modes/diff_mode.go
```diff
@@ -124,9 +126,24 @@ func nullOrValue(value interface{}) string {
...
+		shouldContinue := true
+		if len(headersWhitelist) > 0 {
+			shouldContinue = false
+		}
+		for _, header := range headersWhitelist {
+			if k == header || header == "*" {
+				shouldContinue = true
+			}
+		}
+		if !shouldContinue {
```
tommysitu: The check suggests `should not continue` but it does `continue` to go to the next iteration. Maybe `shouldDoDiff`or `isWhitelistedForDiffing` would be a better name?

tommysitu:
Adding a blacklist option, and support blacklisting certain fields in the response body would be very useful too.

tommysitu:
Actually, I think the diff should by default diff everything, and you set a flag to control what should be excluded would make more sense

benjih:
Updated branch to be a blacklist instead of a whitelist. Default behaviour should now be unchanged, resulting in diffs including headers unless they are filtered out with.

```
hoverctl mode diff --all-headers
```

PR #760, opened 2018-07-31
Correct Response Example, and Put Body

Warfront1:
Perhaps other API is impacted but in particular the mode PUT does not work as displayed in the example docs.
I've corrected that, and also included a realistic GET response


tommysitu:
Thank you! @Warfront1

Issue #764, opened 2018-08-21
Capturing 3 or more sequences leads to simulation problems


john4744:
I start hoverfly, put it in capture mode with the `--stateful` option. Then issue the following requests:
```
#These requests capture with sequence state = sequence:1
curl -k -x http://localhost:8500 http://worldclockapi.com/api/json/utc/now
curl -k -x http://localhost:8500 http://worldclockapi.com/api/json/utc/now

#These requests capture with sequence state = sequence:2
curl -k -x http://localhost:8500 http://worldclockapi.com/api/json/est/now
curl -k -x http://localhost:8500 http://worldclockapi.com/api/json/est/now

#These requests capture with sequence state = sequence:1
curl -k -x http://localhost:8500 http://worldclockapi.com/api/json/pst/now
curl -k -x http://localhost:8500 http://worldclockapi.com/api/json/pst/now
```
I save this simulation to disk, then put hoverfly in simulate mode and import my simulation. When I issue the same set of requests again, I notice that my first request to `http://worldclockapi.com/api/json/pst/now` returns the second response instead of the first response.

I believe this is happening because it shares the same sequence state key as the first sequence that was captured.
<details><summary>Exported simulation json</summary><p>

```json
{
	"data": {
		"pairs": [
			{
				"request": {
					"path": [
						{
							"matcher": "exact",
							"value": "/api/json/utc/now"
						}
					],
					"method": [
						{
							"matcher": "exact",
							"value": "GET"
						}
					],
					"destination": [
						{
							"matcher": "exact",
							"value": "worldclockapi.com"
						}
					],
					"scheme": [
						{
							"matcher": "exact",
							"value": "http"
						}
					],
					"body": [
						{
							"matcher": "exact",
							"value": ""
						}
					],
					"query": {},
					"requiresState": {
						"sequence:1": "1"
					}
				},
				"response": {
					"status": 200,
					"body": "{\"$id\":\"1\",\"currentDateTime\":\"2018-08-21T22:35Z\",\"utcOffset\":\"00:00:00\",\"isDayLightSavingsTime\":false,\"dayOfTheWeek\":\"Tuesday\",\"timeZoneName\":\"UTC\",\"currentFileTime\":131793645274283656,\"ordinalDate\":\"2018-233\",\"serviceResponse\":null}",
					"encodedBody": false,
					"headers": {
						"Cache-Control": [
							"no-cache"
						],
						"Content-Length": [
							"233"
						],
						"Content-Type": [
[...]

PR #765, opened 2018-08-22
Maintain existing state instead of overwriting

john4744:
I raised this issue yesterday (https://github.com/SpectoLabs/hoverfly/issues/764), this branch seems to fix it.
After some debugging it seemed like the entire state was being overwritten each time a new sequence is added. I'm new to Hoverfly and new to Go, so feedback is very welcome 😄 
I ran through the existing tests and it didn't seem to cause any new failures.


benjih:
Excellent spot, very much appreciated @john4744! Thank you!

Issue #762, opened 2018-08-16
FR: Support using upstream-proxy conditionally


ohenak:
Problem: In my company, the proxy for accessing external world is restricted to only access external websites but not internal. However, there are use cases that we need Hoverfly to be able to access both external and internal websites.

Is there a way to have Hoverfly use the upstream-proxy based on certain condition, something like exclusion list or proxy auto-config (PAC) file?

benjih:
Hey @ohenak,

I am currently in the process of implementing this. It should be available in v0.17.4.

tommysitu:
I believe @benjih has implemented it in v0.17.4

Issue #583, opened 2017-06-08
Run hoverfly using automatic configuration script


kallolghose:
Hi,

I am using hoverfly as proxy server to capture responses in a WebPage, I am sitting behind a corporate automatic configuration script, so when I run hoverfly am getting error **(connectex: A connection attempt failed because the connected party did not properly respond)** open as I am not able set the automatic configuration script to my proxy server. 

So, is there a way to start the hoverfly using automatic configuration script(.dat) file.

Thanks in advance !!

benjih:
Hey @klloghose

Are the requests you are trying to capture from the webpage using web sockets?

kallolghose:
Hi @benjih 

Yes i am trying to capture the request from webpage but not using web sockets. The webpages are behind a corporate proxy which is implemented using Proxy Auto-Config(PAC) file. 
Hoverfly uses upstream-proxy but we need to use the PAC file. Is there a way to do this ?
OR 
Is there a way to send some requests through the upstream-proxy and some without using any proxy?

Thanks !!

JohnFDavenport:
Closing this issue as it has gone stale. Reopen if needed! It seems someone intended to make it hard for tools such as Hoverfly here. Hoverfly is primarily an API Simulation or Service Virtualization tool - we'd need to see more demand for use-cases such as this - or maybe a PR!

PR #766, opened 2018-08-23
Implement Hoverfly configuration by PAC file

benjih:
I have added a new endpoint to Hoverfly called `/v2/api/hoverfly/pac`. A user can GET or PUT a PAC file from/to this endpoint. Putting a PAC file on this endpoint will allow Hoverfly to dynamically set the upstream proxy based on the result of the PAC file. This should fulfill issue #762 and older issue #583.

PAC files are Javascript functions that are usually executed by a client like a web browser.
```
function FindProxyForURL(url, host) {
	// our local URLs from the domains below example.com don't need a proxy:
	if (shExpMatch(host, "*.example.com"))
	{
		return "DIRECT";
	}

	// URLs within this network are accessed through
	// port 8080 on fastproxy.example.com:
	if (isInNet(host, "10.0.0.0", "255.255.248.0"))
	{
		return "PROXY fastproxy.example.com:8080";
	}

	// All other requests go through port 8080 of proxy.example.com.
	// should that fail to respond, go directly to the WWW:
	return "PROXY proxy.example.com:8080; DIRECT";
}
```
I have achieved this by using [jackwakefield/gopac](https://github.com/jackwakefield/gopac), which uses [github.com/robertkrimen/otto](robertkrimen/otto), A Javascript interpreter written in Golang. If a PAC file is provided, it will get parsed and executed in by the interpreter and then with the result, will configure a HTTP client correctly.

There are functional tests in this PR proving Hoverfly retrieving responses from two different Hoverfly instances running in simulate mode, with the routing between the two instances determined by a PAC file.


benjih:
Forgot to mention, as I needed to pull new dependency in, things obviously went sideways, hence all the vendor changes.

Issue #767, opened 2018-08-24
Hoverfly crash when processing simultaneous requests


john4744:
I'm running in simulation mode (though I've hit a similar crash in capture mode too). The simulation is stateful. Here's the error message: `fatal error: concurrent map read and map write`.
It seems like access to the State map isn't synchronized.

<details><summary>Here is the full dump</summary><p>

```
fatal error: concurrent map read and map write

goroutine 72 [running]:
runtime.throw(0xac457f, 0x21)
 C:/Go/src/runtime/panic.go:616 +0x88 fp=0xc042faaad8 sp=0xc042faaab8 pc=0x42c6b8
runtime.mapaccess1_faststr(0x9f9ae0, 0xc04274ff80, 0xc0426955b0, 0xb, 0xc042982698)
 C:/Go/src/runtime/hashmap_fast.go:181 +0x428 fp=0xc042faab48 sp=0xc042faaad8 pc=0x40a358
github.com/SpectoLabs/hoverfly/core/matching.StateMatcher(0xc042004e50, 0xc042260450, 0xab123a)
 C:/Users/john4744/go/src/github.com/SpectoLabs/hoverfly/core/matching/state_matcher.go:21 +0x160 fp=0xc042faac18 sp=0xc042faab48 pc=0x84ab50
github.com/SpectoLabs/hoverfly/core/matching.MatchingStrategyRunner(0xc04347c174, 0x3d, 0xc04347c150, 0x3, 0xc04347c15b, 0x19, 0xc04347c154, 0x4, 0xc042c3a6f0, 0x0, ...)
 C:/Users/john4744/go/src/github.com/SpectoLabs/hoverfly/core/matching/matching_strategy_runner.go:36 +0x481 fp=0xc042fab100 sp=0xc042faac18 pc=0x84a431
github.com/SpectoLabs/hoverfly/core/matching.Match(0xab2f32, 0x9, 0xc04347c174, 0x3d, 0xc04347c150, 0x3, 0xc04347c15b, 0x19, 0xc04347c154, 0x4, ...)
 C:/Users/john4744/go/src/github.com/SpectoLabs/hoverfly/core/matching/matcher.go:12 +0x192 fp=0xc042fab1b0 sp=0xc042fab100 pc=0x849f82
[...]

PR #769, opened 2018-09-06
Protect state from concurrent access

john4744:
Attempting to fix https://github.com/SpectoLabs/hoverfly/issues/767 by adding a mutex to protect access to the State's map. I've run the existing tests and it didn't seem to cause any new failures.


benjih:
Fantastic work again @john4744. We really appreciate your continued effort to help us in supporting Hoverfly!

PR #771, opened 2018-09-19
Add mutex to protect access to Simulation.matchingPairs

john4744:
When using Hoverfly in stateful capture mode, I was hitting it with the same request from multiple threads and noticed that the captured simulation had gaps and duplicates in the sequence state numbering. For instance, I saw 2 identical requests saved with `"sequence:1": "11"`, and then another request saved with `"sequence:1": "13"`. It didn't appear to save one with `"sequence:1": "12"`.
After a bit of debugging, it seemed that maybe 2 threads were concurrently entering `Simulation. AddPairInSequence` and causing some havoc.
I added the mutex locking below, and now I haven't hit that problem again. Let me know what you think.

### DIFF
--- a/core/models/simulation.go
+++ b/core/models/simulation.go
@@ -4,6 +4,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/SpectoLabs/hoverfly/core/state"
 )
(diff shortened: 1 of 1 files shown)


JohnFDavenport:
Thanks @john4744 - much appreciated - we'll have a look at that as soon as we can

JohnFDavenport:
A question @john4744 - I'm very interested as you seem to be using Hoverfly under load in Capture mode. Is that just a consequence of how you are capturing data? Reason I'm asking is we clearly haven't anticipated that and that means there's a gap in our understanding. Feel free to contact me using the address in my profile.

Issue #201, opened 2016-06-01
Implement two way SSL


mogronalol:
Proposed solution for this is to allow Hoverfly to be configurable with a client certificate to use with two way SSL.  

It will then be up to the end user to decide on whether to generate a new client certificate for Hoverfly, or re-use an existing client certificate.  The former is most likely to be used as it means avoiding sharing of private keys between existing clients and Hoverfly, but also means that the server will have to be configured to trust the new certificate used by Hoverfly.

sarathchand09:
Could you please tell me how to do the ssl config in hoverfly, how do we configure client certificates in hoverfly for oneway or 2 way ssl

mogronalol:
SSL setup information is available in the docs [here](https://spectolabs.gitbooks.io/hoverfly/content/usage/certificate_management.html)

sarathchand09:
Thanks, is two way-ssl already implemented? or still under development for hoverfly

Osipion:
Hey. I'd like an update on this - is it implemented yet @mogronalol? Looks like it was removed from the delivery milestones for last year. However, we are also unable to use Hoverfly without two-way SSL

mogronalol:
Unfortunately, this has fallen behind due to the development of our cloud offering. It can be put back on our roadmap but it will likely not be implemented in the immediate future.

JohnFDavenport:
@Osipion @sarathchand09 We'd be interested in learning more about your requirements.

If you prefer to, feel free to contact us directly.

Osipion:
@JohnFDavenport OK, so I can't talk about specific tech beyond the development OS (OS X), Docker, and Hoverfly, but I might have an architecture something like this
```
        [Device]
           | 1 
         [API]
    |2     |2      |2
   [A]    [B]     [C]
```
Where A, B and C are HTTP services the API consumes. All connections (1 & 2) use two-way pinned SSL (I have the cert and key). The server expects clients to present a certificate, and the client expects the server to present a pinned certificate.

I have two main use cases:

### Dev:

[...]

PR #773, opened 2018-10-02
Add options to support capture request with 2way ssl authentication

xiang-zhu:
Add options to make hoverfly support capture request using client SSL authentication. I think it's the same object as described in https://github.com/SpectoLabs/hoverfly/issues/201

With the new options, the request can be set to use client ssl cert and keys to authenticate if the request host matches provided pattern., and won't affect other requests. The default value of these options won't have any impact to exisiting logic, so it's a non breaking change.

We've been using the forked binary to capture data with client SSL authentication in our test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,10 @@ requests.db
 *.iml
 target
 
+# Vim
+*.swp
+*.swo
+
 core/cmd/hoverfly/hoverfly
 hoverctl/hoverctl
 functional-tests/core/bin/hoverfly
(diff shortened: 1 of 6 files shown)


tommysitu:
@xiang-zhu You rocks! This is certainly a sought-after feature by hoverfly users. I am hoping to do some testing around your PR and will get back to you.

xiang-zhu:
Thank you for testing this @tommysitu we really like what hoverfly provides and hope it's getting better.

PR #789, opened 2018-11-23
added log level flag

podtserkovskiy:
Log level `info` causes degradation of performance.


tommysitu:
Looks good. The `info` logging has a huge impact on the performance.

PR #794, opened 2019-01-16
Join multi-value query with semicolon when saving to a simulation

tommysitu:


JohnFDavenport:
A small change and a good test. I'm fine with this.

PR #797, opened 2019-01-17
Fix proxy tunnel request

tommysitu:


JohnFDavenport:
Nice that you spotted a safer way to get the hostname. Test looks good to.

PR #799, opened 2019-02-02
Correct ineffectual assignment

gliptak:


tommysitu:
Thank you. It reminds me that we are overdue for some go linting.

PR #803, opened 2019-02-08
Default remote target protocol should be HTTP

tommysitu:
When the scheme is missing in the host flag for a remote target, it should just default to HTTP. This is also common among other tools like `curl`. 

Users should override the default settings themselves if they are connecting to a remote Hoverfly Admin API runs on HTTPS.

### DIFF
--- a/hoverctl/wrapper/hoverfly.go
+++ b/hoverctl/wrapper/hoverfly.go
@@ -14,9 +14,7 @@ import (
 	"time"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/SpectoLabs/hoverfly/core/handlers"
 	"github.com/SpectoLabs/hoverfly/core/handlers/v2"
-	"github.com/SpectoLabs/hoverfly/core/util"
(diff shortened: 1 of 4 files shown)


hoverctl/wrapper/hoverfly.go
```diff
@@ -163,11 +163,11 @@ func Login(target configuration.Target, username, password string) (string, erro
 
 func BuildURL(target configuration.Target, endpoint string) string {
 	if !strings.HasPrefix(target.Host, "http://") && !strings.HasPrefix(target.Host, "https://") {
-		if IsLocal(target.Host) {
+		//if IsLocal(target.Host) {
 			return fmt.Sprintf("http://%v:%v%v", target.Host, target.AdminPort, endpoint)
-		} else {
-			return fmt.Sprintf("https://%v:%v%v", target.Host, target.AdminPort, endpoint)
-		}
+		//} else {
```
benjih: No commented out code.

tommysitu: Fixed

PR #804, opened 2019-02-08
Caching parsed response template

tommysitu:
This is recommended by Raymond handlebar library https://github.com/aymerick/raymond#correct-usage to avoid parsing template on every request. 

Benchmark results before and after:
Before
```
BenchmarkProcessRequest/Templated_simulation-8            300000             98836 ns/op
```
After
```
BenchmarkProcessRequest/Templated_simulation-8           3000000             11367 ns/op
```


tommysitu:
@benjih this PR is up for grabs as well.

Issue #802, opened 2019-02-08
Performance issue with large response body containing templating functions


tommysitu:
An issue was reported regarding load testing hoverfly which uses excessive CPU.  This happened when simulation contains large response body (about 800KB) and templating functions.

PR #810, opened 2019-02-15
#802 Only set content-length header if transfer-encoding header is ab…

tommysitu:
…sent, add more e2e test around response headers


tommysitu:
For HTTPS proxying, GoProxy always set Transfer-Encoding header and remove existing Content-Length header as the safest way to write out the response body. Therefore a wrong content length does not matter (for example when templating or middleware modified the response body)

For HTTP proxying and webserver mode, a Content-Length header is only calculated if the body is not null, and neither Content-Length or Transfer-Encoding exists. Therefore if your simulation contains wrong Content-Length header value, Hoverfly will give you a warning when importing and your HTTP client might fail when reading the response body.

PR #813, opened 2019-02-21
Update capturingsequences.rst

JohnFDavenport:
Documentation doesn't explain that it's stateful services rather than sequences. Hoverfly does the capture in a sequence as 90% of the time that's enough or close enough that it gives a good result.


tommysitu:
LGTM

PR #824, opened 2019-03-14
rename files so Hoverfly can be depended upon as a go module

DanielJonesEB:
You've got a couple of files that appear to have two full-stops (`..`) in their names erroneously. This prevents projects from being able to depend on Hoverfly as a Go Module, because [`go` will reject any file paths containing `..`](https://github.com/golang/go/issues/27299) (and also `:` apparently):

```
go: extracting github.com/SpectoLabs/hoverfly v0.17.7
-> unzip /Users/deejay/go/pkg/mod/cache/download/github.com/!specto!labs/hoverfly/@v/v0.17.7.zip: malformed file path "core/handlers/v2/hoverfly_upstream_proxy_handler..go": double dot
```


tommysitu:
Thank you for spotting it out!

PR #831, opened 2019-04-10
adding partial matching for json

dilitvinov:
This PR adds a partial matching feature for JSON and tests for it.

### DIFF
--- a/core/matching/matchers/json_partial_match.go
+++ b/core/matching/matchers/json_partial_match.go
@@ -1,0 +1,43 @@
+package matchers
+
+import "encoding/json"
+
+var JsonPartial = "jsonpartial"
+
+func JsonPartialMatch(match interface{}, toMatch string) bool {
+	var expected, actual map[string]interface{}
+	matchString, ok := match.(string)
+	if !ok {
+		return false
+	}
+	err0 := json.Unmarshal([]byte(matchString), &expected)
+	err1 := json.Unmarshal([]byte(toMatch), &actual)
+	if err0 != nil || err1 != nil {
+		return false
+	}
+	return isMapContainsMap(expected, actual)
+}
+
+func isMapContainsMap(match, toMatch interface{}) bool {
+	var expected, actual map[string]interface{}
+	expected, ok0 := match.(map[string]interface{})
+	actual, ok1 := toMatch.(map[string]interface{})
+	if !ok0 || !ok1 {
+		return false
+	}
+
+	for key, val := range expected {
+		if innerMap, ok := val.(map[string]interface{}); ok {
+			result := isMapContainsMap(innerMap, actual[key])
+			if !result {
+				return false
+			}
+		} else {
+			_, exist := actual[key]
+			if !exist {
+				return false
+			}
+		}
+	}
+	return true
+}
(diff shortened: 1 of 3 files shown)


tommysitu:
Nice PR! From the tests, I can see that this feature allow JSON match to ignore unknown fields if they are not defined in the matcher. I am wondering if it is capable to match a  JSON node that is located in any nested level of a JSON map like this:  

The request matcher value is: 
```
{
	"fieldA": "valueA"
}
```
and the actual JSON to match is: 
```
{
	"test": {
		"json": true,
		"minified": true,
		"someObject": {
			"fieldA": "valueA"
		}
	}
}
```

dilitvinov:
OK, I'll do that

tommysitu:
No pressure, I was just checking. What you have got should be good enough. For the above scenario, someone can define this matcher value to match using your implementation: 

```
{
	"test": {
		"someObject": {
			"fieldA": "valueA"
		}
	}
}
```

dilitvinov:
Yes, but we may not know about parent nodes for eg. It sounds like a good case and worth to be implemented.

dilitvinov:
I've done work, check it out, please

tommysitu:
Thanks! Will take a look as soon as possible.

core/matching/matchers/json_partial_match.go
```diff
@@ -0,0 +1,132 @@
+package matchers
+
+import (
+	"encoding/json"
+)
+
+var JsonPartial = "jsonpartial"
+
+func JsonPartialMatch(match interface{}, toMatch string) bool {
+	var expected, actual map[string]interface{}
```
tommysitu: This may not work if the match or toMatch JSON is an array, and starts with a square bracket. Can you confirm?

tommysitu:
Can you squash the last few commits into one called `deep search added` please? we try to keep the git history clean.

core/matching/matchers/json_partial_match.go
```diff
@@ -0,0 +1,132 @@
+package matchers
+
+import (
+	"encoding/json"
+)
+
+var JsonPartial = "jsonpartial"
```
tommysitu: Can you change this to use camelCase please? Just a convention we try to follow for the simulation file.

core/matching/matchers/json_partial_match.go
dilitvinov: You right, It doesn't.
I can make this ability, but I'm not sure that described data structure is a valid json.

core/matching/matchers/json_partial_match.go
dilitvinov: var JsonPartial = "jsonPartial"

This way, right?

dilitvinov: I've chosen this way cuz it looks like jsonpath example

var JsonPath = "jsonpath"

dilitvinov: Anyway, i changed name

dilitvinov:
> Can you squash the last few commits into one called `deep search added` please? we try to keep the git history clean.

To be honest, I'm not a git-wizard. So could you please squash commits on pr merge level?
Something like this https://help.github.com/en/articles/merging-a-pull-request

tommysitu:
Looks good. Thank you!

PR #838, opened 2019-04-29
Roll back goproxy update which impacts performance

tommysitu:


JohnFDavenport:
This is rolling back to when Hoverfly was in a happier state wrt performance.

PR #844, opened 2019-06-03
Import multiple simulations

tommysitu:


tommysitu:
> simulation.go line 19
> "add [paths to simulations]" is confusing as the user may think _I add the paths here but where do I add the simulations_?. Perhaps just say "add [simulations]" as the long description will tell you each simulation has a path and filename.
> 
> simulation.go line 22 suggested text:
> Adds one or more simulation files to Hoverfly to
> the existing simulation data.
> Any request/response pairs that have an identical request to those in the
> existing data will be discarded with a warning message.
> 
> You may provide an absolute or relative paths to each simulation file.

Updated

PR #858, opened 2019-07-03
Fix slow import

tommysitu:
With this PR, you can disable duplicate pair check on import, by starting hoverfly with: 
```
hoverfly -no-import-check
```

or 
```
hoverctl start --no-import-check
```


aosavitski:
Works great with "--no-import-check"!

Issue #855, opened 2019-06-27
Poor performance when simulation has a lot of pairs captured in stateful mode


aosavitski:
Simulations captured with stateful mode are very slow, because copy of the state.State map is done in /hoverfly/core/matching/state_matcher.go and /hoverfly/core/matching/strongest_match_strategy_test.go

`state_matcher.go`
`currentState.RWMutex.RLock()`
`copyState := util.CopyMap(currentState.State)`
`currentState.RWMutex.RUnlock()`

`strongest_match_strategy.go`
`state.RWMutex.RLock()`
`copyState := util.CopyMap(state.State)`
`state.RWMutex.RUnlock()`

And problem is not in locking added in [hoverfly#767](https://github.com/SpectoLabs/hoverfly/issues/767), but in coping this map in the loop. In state_matcher.go we are interested in just one pair from this map, but we are coping whole map for each pair... In strongest_match_strategy.go we put this copy in the ClosestMiss for each miss with the same or better miss score. Why? Just to show this whole map in the log at the end? I think it could be improved. As a quick an dirty fix I can offer this solution:
We are doing this copyMap only once in matching_strategy_runner.go, before we start loop. And then StateMatcher and PostMatching will have as argument `copyState map[string]string` instead of `currentState *state.State`.

aosavitski:
> @tommysitu 
> @aosavitski maybe the locking and map copy can be replaced by syncmap >https://godoc.org/golang.org/x/sync/syncmap. Can you try it out please? You’ve got the perfect >environment to test out this fix. A PR will be really welcome.

@tommysitu [documentation](https://tip.golang.org/pkg/sync/#Map) saying: `The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination`
state.State is already an API on top of original map with synchronisation. So may be better is not to replace it, but just stop accessing to the map data directly?
So in state_matcher.go instead of creating a copy and accessing value with copyMap[key] it could be:
```
currentStateValue, ok := currentState.GetState(key)
if !ok {
	matched = false
}
if currentStateValue != value {
	matched = false
} else {
	score++
}
```
instead of current:
```
currentState.RWMutex.RLock()
copyState := util.CopyMap(currentState.State)
currentState.RWMutex.RUnlock()
for key, value := range requiredState {
if _, ok := copyState[key]; !ok {
	matched = false
}
if copyState[key] != value {
	matched = false
} else {
	score++
}
```
[...]

PR #859, opened 2019-07-03
Fix slow stateful mode #855

aosavitski:
With this PR you will a better simulation performance in case you have data captured in stateful mode

### DIFF
--- a/core/matching/matching_strategy_runner.go
+++ b/core/matching/matching_strategy_runner.go
@@ -3,16 +3,20 @@ package matching
 import (
 	"github.com/SpectoLabs/hoverfly/core/models"
 	"github.com/SpectoLabs/hoverfly/core/state"
+	"github.com/SpectoLabs/hoverfly/core/util"
 )
 
 type MatchingStrategy interface {
 	PreMatching()
 	Matching(*FieldMatch, string)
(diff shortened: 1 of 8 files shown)


core/matching/matching_strategy_runner.go
```diff
@@ -35,7 +39,7 @@ func MatchingStrategyRunner(req models.RequestDetails, webserver bool, simulatio
 
 		strategy.Matching(StateMatcher(state, requestMatcher.RequiresState), "state")
```
tommysitu: Good optimization to remove the copy map from the loop. However I want to drop a note about the race condition here. The state could be altered by another thread at any one point during the matching, and the matching error message could be saying Hoverfly has the state of `foo=bar` and the required state is also `foo=bar` which is contradicting.

I think the whole matching operation should be synchronized to avoid the race condition, it's not safe to let other thread to modify the states while matching is in progress. I can't tell what the performance implication is, and we should fix it at some point.

aosavitski: We can use copyState map (that we create before loop starts) as the parameter in StateMatcher.

tommysitu: That's true. Would you like to submit another PR?

aosavitski: Sure

PR #860, opened 2019-07-07
Update templating.rst

jiimaho:


tommysitu:
Thank you for spotting out the error.

PR #863, opened 2019-07-09
#855 Use copyState map as the parameter in StateMatcher function

aosavitski:
Use copyState map as the parameter in StateMatcher function to avoid race condition


tommysitu:
LGTM

PR #875, opened 2019-08-20
Realised handling templating on headers response

kulaginds:
I made handling templates in response headers.
I update `core/hoverfly_funcs_test.go#Test_Hoverfly_GetResponse_WillCacheTemplateIfNotInCache` test: add header `X-Image-Id` with template `{{ randomInteger }}`.

How I test it:
1. Make local docker image: `docker build -t kulaginds/hoverfly`.
2. Run docker image: `docker run -ti --rm -p 8888:8888 -p 8500:8500 kulaginds/hoverfly:latest`
3. Put [test.json](https://gist.github.com/kulaginds/893d7d8fff142d9a218f1b4a3438e35c) to hoverfly.
4. Execute `curl -vvv -x localhost:8500 http://google.com/hello`.
Expected random integer in header `X-Image-Id`, actual `3551771716390149526`.

### DIFF
--- a/core/hoverfly_funcs.go
+++ b/core/hoverfly_funcs.go
@@ -86,23 +86,20 @@ func (hf *Hoverfly) GetResponse(requestDetails models.RequestDetails) (*models.R
 	// as we use the current state in templates
 	if response.Templated == true {
 
-		var template *raymond.Template
-		if cachedResponse != nil && cachedResponse.ResponseTemplate != nil {
-			template = cachedResponse.ResponseTemplate
-		} else {
-			// Parse and cache the template
-			template, _ = hf.templator.ParseTemplate(response.Body)
-			if cachedResponse != nil {
-				cachedResponse.ResponseTemplate = template
-			}
-		}
-
-		responseBody, err :=  hf.templator.RenderTemplate(template, &requestDetails, hf.state.State)
+		responseBody, err := hf.applyBodyTemplating(&requestDetails, &response, cachedResponse)
 
 		if err == nil {
 			response.Body = responseBody
 		} else {
-			log.Warnf("Failed to render response template: %s", err.Error())
+			log.Warnf("Failed to applying body templating: %s", err.Error())
+		}
+
+		responseHeaders, err := hf.applyHeadersTemplating(&requestDetails, &response, cachedResponse)
+
+		if err == nil {
+			response.Headers = responseHeaders
+		} else {
+			log.Warnf("Failed to applying headers templating: %s", err.Error())
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


core/hoverfly_funcs.go
```diff
@@ -117,6 +114,65 @@ func (hf *Hoverfly) GetResponse(requestDetails models.RequestDetails) (*models.R
...
+
+	return hf.templator.RenderTemplate(template, requestDetails, hf.state.State)
+}
+
+func (hf *Hoverfly) applyHeadersTemplating(requestDetails *models.RequestDetails, response *models.ResponseDetails, cachedResponse *models.CachedResponse) (map[string][]string, error) {
+	var headersTemplates map[string][]*raymond.Template
+	if cachedResponse != nil && cachedResponse.ResponseHeadersTemplates != nil {
+		headersTemplates = cachedResponse.ResponseHeadersTemplates
+	} else {
+		var header []*raymond.Template
```
tommysitu: For brevity, can you remove this line and use short variable declaration like: `header := make([]*raymond.Template, len(v))`?

core/hoverfly_funcs.go
```diff
@@ -117,6 +114,65 @@ func (hf *Hoverfly) GetResponse(requestDetails models.RequestDetails) (*models.R
...
+			headersTemplates[k] = header
+		}
+
+		if cachedResponse != nil {
+			cachedResponse.ResponseHeadersTemplates = headersTemplates
+		}
+	}
+
+	var (
+		header []string
```
tommysitu: this too. `header := make([]string, len(v))`

core/hoverfly_funcs_test.go
```diff
@@ -336,6 +341,7 @@ func Test_Hoverfly_GetResponse_WillCacheTemplateIfNotInCache(t *testing.T) {
 	Expect(found).To(BeTrue())
 
 	Expect(cachedRequestResponsePair.(*models.CachedResponse).MatchingPair.Response.Body).To(Equal("{{ randomUuid }}"))
+	Expect(cachedRequestResponsePair.(*models.CachedResponse).MatchingPair.Response.Headers["X-Image-Id"][0]).To(Equal("{{ randomInteger }}"))
```
tommysitu: Perhaps you can try adding a test in this test suite to check if the header template is rendered properly? Or you can add a functional test here: https://github.com/SpectoLabs/hoverfly/blob/master/functional-tests/core/ft_templated_response_test.go
Whichever you prefer.

core/hoverfly_funcs.go
kulaginds: It good proposal, but `response.Headers` below has signature `map[string][]string`. Every header has different length slice of strings.

core/hoverfly_funcs.go
kulaginds: This too every header has different length slice of strings.

core/hoverfly_funcs_test.go
kulaginds: Add unit-test `core.Test_Hoverfly_GetResponse_WillCacheHeaderTemplateIfNotInCache` and functional-test `randomInteger in header` in file `functional-tests/core/ft_templated_response_test.go`

core/hoverfly_funcs.go
```diff
@@ -117,6 +114,65 @@ func (hf *Hoverfly) GetResponse(requestDetails models.RequestDetails) (*models.R
...
+		err error
+	)
+	headers := map[string][]string{}
+
+	// Render headers templates
+	for k, v := range headersTemplates {
+		header = make([]string, len(v))
```
tommysitu: I meant you can do `header := make([]string, len(v))` here, it will create a slice with variable length, and you don't need the `var` declaration in line 155.  Anyway, it's not a big deal.

tommysitu:
@kulaginds this enhancement is released in v1.1.2. Sorry for the delay!

PR #884, opened 2019-11-18
Fix race condition with simulatenously appending new entries to journal

ttiurani:
This fixes the issue with missing journal entries described here: 

https://github.com/SpectoLabs/hoverfly-java/issues/225

The fix might not be perfect, but where previously 5 repetitions caused the error, with this change it can't be replicated in over 100 repetitions.

NOTE: I don't know how to write Go so putting the mutex to the Journal struct might be really dumb. But I'm willing to make this sensible given comments.

### DIFF
--- a/core/journal/journal.go
+++ b/core/journal/journal.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"time"
+	"sync"
 
 	sorting "sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


core/journal/journal.go
```diff
@@ -27,6 +28,7 @@ type JournalEntry struct {
 type Journal struct {
 	entries    []JournalEntry
 	EntryLimit int
+	mutex      sync.Mutex
```
tommysitu: `sync.RWMutex` is preferred as we need to synchronize the reading as well. With `sync.Mutex` readers need to wait for each other.

tommysitu: Actually forget about the read lock, as too many read locking could block writes and in turns block proxying with current implementation. journal write need to made async via a channel or some other way first to avoid this issue

Issue #815, opened 2019-02-24
Allow responses being read from a file instead of having to write them in the simulation JSON


BrainStone:
**Is your feature request related to a problem? Please describe.**
Currently when running a simulation you have to specify the response bodies in the simulation JSON directly. This makes formats that use double quotes or new lines (like JSON) pretty much unreadable. And when using binary formats like images, it's essentially impossible to know which image is which without converting the base64 encoded image back first.  
Additionally having the response body in the simulation JSON can really blow up the file size.

**Describe the solution you'd like**
I think parallel to the `body` property, there should be a `bodyFile` property where you can specify a file from which the body can be read from. Relative paths should be relative to the simulation JSON file itself.

This might also allow to lazily load large files to improve general performance. As it may make sense to create a large set of response definitions in one file, but not all responses are requested and therefore not all response bodies would need to be loaded into memory.

**Describe alternatives you've considered**
One alternative that has been pointed out is using middleware. Though I think it's really way over the top to essentially write a small program to solve such a trivial and potentially common problem. Additionally you always run into the issue of platform dependence when using such a solution.  
If there was a premade, complete and platform independent script/program that loads responses from files, while still allowing direct response bodies in the simulation JSON, then that would be acceptable too, though I really think it would be benefitial to have that feature built in Hoverfly itself directly

JohnFDavenport:
Makes sense as JSON is difficult to work with and the bodies would be separately managed

crorodriguezro:
Hi, any update on this?

tommysitu:
Please try this out in [v1.3.0](https://github.com/SpectoLabs/hoverfly/releases/tag/v1.3.0) It might need some more refinement.

Issue #807, opened 2019-02-13
Per-stub delays


tommysitu:
**Is your feature request related to a problem? Please describe.**
Delays setting in `globalActions` is not flexible enough, you can only filter request by `urlPattern` or `httpMethod`. You can't filter by request body or headers for example. The main issue is that when you need granular control of delays, you have to write a lot of urlPatterns while requestMatcher already does it for you. 

**Describe the solution you'd like**
You should be able to add delay to individual request-response pair (stub) directly. 
For fixed delay:

```
"response": {
	"status": 200,
	"body": "slow response",
        "fixedDelay": 1000
}
```

For lognormal delay: 

```
"response": {
	"status": 200,
	"body": "slow response",
        "logNormalDelay": {
               "min": 100,
               "max": 10000,
               "mean": 5000,
               "median": 500
            }
}
```

Local delays (per-stub delays) should override global delays if they're happened to match on the same request. Multiple delays definitions on the same level should be additive.

UnasZole:
I have a similar requirement, where I need my virtual service to reproduce the real response time observed on the real service for each different request.

An ideal solution would be the following : 
- Set a delay for each stub
- Allow setting a global delayMultiplier (0 = respond immediately for all requests; 1 = respond with the specified delay)

ns3777k:
@tommysitu I think I'm gonna implement this. Only simulate and spy modes should support this? What did you mean by`Multiple delays definitions on the same level should be additive.`' ?

tommysitu:
That would be very helpful! Feel free to have a go.

1. The delay is applied here: https://github.com/SpectoLabs/hoverfly/blob/9708a10025b41cb68e01f43dd45e9f6d3cfc0ca7/core/hoverfly.go#L209

so it supports any mode except `capture` 

2. Exactly. If the user provides both delays,  the actual delay will be the total. 
So the following is supported: 
```
"response": {
	"status": 200,
	"body": "slow response",
        "fixedDelay": 1000,
        "logNormalDelay": {
               "min": 100,
               "max": 10000,
               "mean": 5000,
               "median": 500
            }
}
```

ns3777k:
Hmm. Taken `SynthesizeMode` for example. It doesn't grab a response from the storage but instead it calls a middleware that returns the response. So in this case, a middleware should be able to return `fixedDelay` and `logNormalDelay`?

[...]

Issue #761, opened 2018-08-03
Matcher Chaining


tommysitu:
With matcher chaining, one can use JSONPath to get a JSON node, and use other matcher to compare its value. For example: 
```
{
  body: [
    {
      matcher: "jsonpath"
      value: “$.user.id"
      doMatch: {
        matcher: “exact”,
        value: “1"
    }
  ]
}
```
This removes the stress of composing and testing complex  jsonpath/xpath expressions, which may not be supported by the current jsonpath/xpath library hoverfly is using. The matcher will be more readable too.

kapishmalik:
@tommysitu can I take this up?

kapishmalik:
@tommysitu  I am currently working on this request.

tommysitu:
Some work has been done here to allow one level of nesting for the `form` matcher: https://github.com/SpectoLabs/hoverfly/pull/1025 you might find it useful @kapishmalik

kapishmalik:
@tommysitu I was thinking to change return type of all matchers from bool to string, bool...i.e. matchedvalue, ismatched or not so that I keep on iterating till doMatch is null. Whenever the value of matcher returns false, I should return false otherwise true. Hope it makes some sense.

PR #893, opened 2020-01-25
add bodyFile to response part of pair

ns3777k:
Hey!
I'm working on https://github.com/SpectoLabs/hoverfly/issues/815 and decided to make a PR so you can tell me if I'm on the right track. Tests pass locally, not sure why they are not run on CI here.

I'll add new tests and update docs after successful review.

Thanks.


tommysitu:
Good stuff! I'll review it this week.

core/handlers/v2/simulation_views_v5.go
```diff
@@ -53,6 +53,7 @@ func (this RequestMatcherResponsePairViewV5) GetResponse() interfaces.Response {
 type ResponseDetailsViewV5 struct {
 	Status           int                 `json:"status"`
 	Body             string              `json:"body"`
+	BodyFile         string              `json:"bodyFile"`
```
tommysitu: Let's create a new struct for schema changes. For this will be `simulation_views_v6.go`. There are other changes I'd like to add to v6

core/models/payload.go
```diff
@@ -275,6 +281,7 @@ func (r *ResponseDetails) ConvertToResponseDetailsViewV5() v2.ResponseDetailsVie
 	return v2.ResponseDetailsViewV5{
```
tommysitu: convert to `ResponseDetailsViewV6`

core/modes/modes.go
```diff
@@ -78,12 +78,22 @@ func ReconstructRequest(pair models.RequestResponsePair) (*http.Request, error)
 }
 
 // ReconstructResponse changes original response with details provided in Constructor Payload.Response
-func ReconstructResponse(request *http.Request, pair models.RequestResponsePair) *http.Response {
+func ReconstructResponse(request *http.Request, pair models.RequestResponsePair) (*http.Response, error) {
```
tommysitu: This function is invoked on every request after the response is fetched from a cache and middleware has been applied. Doing file reading here would be a performance killer.

Ideally the response body should be loaded from the file when it gets imported, it also fails fast if the file is not readable. 

I would suggest putting this logic in `payload.go` -> `NewResponseDetailsFromResponse`

ns3777k:
@tommysitu thanks for review. only 1 question for now. Does making v6 require copy pasting v5, renaming every struct to v6 and replacing v5 with v6 in all tests and other files?

tommysitu:
> @tommysitu thanks for review. only 1 question for now. Does making v6 require copy pasting v5, renaming every struct to v6 and replacing v5 with v6 in all tests and other files?

pretty much. also you need to add a function to upgrade v5  to v6 in `simulation_views_upgrade.go`

ns3777k:
@tommysitu ok, got it, but I'm curios :-) wouldn't it be much easier to maintain the project with only 1 schema and avoiding BC? And in case of BC, release new major version with changelog?

tommysitu:
> @tommysitu ok, got it, but I'm curios :-) wouldn't it be much easier to maintain the project with only 1 schema and avoiding BC? And in case of BC, release new major version with a migration guide (or a tool to upgrade existing files)?

It's a little bit extra work to do but we can ensure all the existing users can benefit from hoverfly upgrades without breaking their tests and CI pipelines. It could spark lots of support issues which is what we try to avoid.

ns3777k:
Ok :-) Here we go again.
Now, if `bodyFile` is not empty, it's read and recorded into `body`. `bodyFile` remains in the response for information.
In `NewResponseDetailsFromResponse` there's no error checking for now, I'll add it tomorrow.
Current tests pass.

tommysitu:
Nice, will check it out tmr

ns3777k:
I've added error handling when bodyFile is missing.

ns3777k:
@tommysitu any chance you can review it soon? :shipit:

tommysitu:
> @tommysitu any chance you can review it soon? :shipit: so i'll get on docs and tests

this weekends for sure

core/import.go
```diff
@@ -133,15 +133,18 @@ func (hf *Hoverfly) ImportFromURL(url string) error {
...
+func (hf *Hoverfly) importRequestResponsePairViews(pairViews []v2.RequestMatcherResponsePairViewV6) (v2.SimulationImportResult, error) {
 	importResult := v2.SimulationImportResult{}
 	initialStates := map[string]string{}
 	if len(pairViews) > 0 {
 		success := 0
 		failed := 0
 		for i, pairView := range pairViews {
 
-			pair := models.NewRequestMatcherResponsePairFromView(&pairView)
+			pair, err := models.NewRequestMatcherResponsePairFromView(&pairView)
```
tommysitu: Why not set the error here , then we don't need to change the signature of `importRequestResponsePairViews` : 
```
pair, err := models.NewRequestMatcherResponsePairFromView(&pairView)
			if err != nil {
				importResult.SetError(err)
				return importResult
			}
```

core/handlers/v2/simulation_views_upgrade.go
```diff
@@ -286,22 +286,101 @@ func upgradeV4(originalSimulation SimulationViewV4) SimulationViewV5 {
...
+			RequestResponsePairs: requestResponsePairs,
 			GlobalActions:        originalSimulation.GlobalActions,
 		},
 		newMetaView(originalSimulation.MetaView),
 	}
 }
 
-func getMatchersFromRequestHeaders(headers map[string][]string) map[string][]MatcherViewV5 {
-	requestHeaders := map[string][]MatcherViewV5{}
+func upgradeV5(originalSimulation SimulationViewV5) SimulationViewV6 {
```
tommysitu: missing unit test for this function

core/handlers/v2/simulation_views_upgrade_test.go
```diff
@@ -80,7 +80,7 @@ func Test_upgradeV1_ReturnsAnUpgradedSimulation(t *testing.T) {
 	Expect(upgradedSimulation.RequestResponsePairs[0].RequestMatcher.DeprecatedQuery[0].Matcher).To(Equal(matchers.Exact))
 	Expect(upgradedSimulation.RequestResponsePairs[0].RequestMatcher.DeprecatedQuery[0].Value).To(Equal("query=query"))
 
-	Expect(upgradedSimulation.RequestResponsePairs[0].RequestMatcher.Headers).To(HaveKeyWithValue("Test", []MatcherViewV5{
+	Expect(upgradedSimulation.RequestResponsePairs[0].RequestMatcher.Headers).To(HaveKeyWithValue("Test", []MatcherViewV6{
```
tommysitu: Can you assert the `bodyFile` is set to empty string after the upgrade please? 

```
Expect(upgradedSimulation.RequestResponsePairs[0].Response.BodyFile).To(Equal(""))
```

tommysitu:
@ns3777k Thank you for your patience. I've finally got a chance to test it thoroughly. Here is what I found:

1. If the body file failed to read, it will wipe the existing simulation, because of this line: 
https://github.com/SpectoLabs/hoverfly/blob/master/core/handlers/v2/simulation_handler.go#L127
I think this is something I need to fix afterwards

2. I couldn't get it to read a file relative to the hoverfly binary or the imported simulation. I think what would be really useful is that the body file path should be relative to the simulation file

3. When both `body` and `bodyFile` exists, Hoverfly should pick one or the other, and prints a warning to the user. I would suggest `body` field takes precedence in this case.

ns3777k:
I have fixed the comments above.

1. Should I fix this? Can you describe the case in more details?

2. Hmm. Can you show me the failed use case? Relative paths are resolved to the cwd, not the binary. If hoverfly was started from another directory, then the cwd would be that another directory. Starting hoverfly with hoverctl does not set the cwd for hoverfly (https://github.com/SpectoLabs/hoverfly/blob/master/hoverctl/wrapper/hoverfly.go#L173). Will setting the cwd there solve the case?

3. For now it picks bodyFile and prints no warning. I'll change that this week.

tommysitu:
Regarding to the comments: 
1. feel free to fix it if you got some ideas. we shouldn't allow a failed import deleting the simulation that was storing in Hoverfly

2. Basically I couldn't get relative path working at all. I've tried putting the file in the same directory as hoverfly binary or the simulation file, or the cwd. None of that works. 

In the issue https://github.com/SpectoLabs/hoverfly/issues/815,  @BrainStone proposed that the relative path could be relative to the Simulation file. I think that's a good idea, as you can keep these body files side-by-side to the simulation where they can be found easily. 

For example, you can store your simulation and body file with this structure:
```
.
+-- simulations 
|   +-- flight_api.json
|   +-- flight_api_responses
    |   +-- 200.json
    |   +-- 400.json

```
and the `bodyFile` in your `flight_api.json` can set to `flight_api_response/200.json` and `flight_api_response/400.json` regardless of where you start your hoverfly or use hoverctl.

3. Thanks for the PR. The fix looks good to me. No need to change the export cmd. As we have given user a warning, we'd expect them to fix the simulation and decide what they want to keep.

ns3777k:
1. Ok, I'll think about it this week
2. Hmmm. Say I wanna use `hoverctl import` or API directly to send a pair with bodyFile in response. The simulation file is located locally (on client) but hoverfly service can be run on a different machine and hoverfly binary won't be able to find files relative to simulation file thay was uploaded from a client machine. Am I missing something here? :confused: 
3. Ok, cool

tommysitu:
> mmm. Say I wanna use hoverctl import or API directly to send a pair with bodyFile in response. The simulation file is located locally (on client) but hoverfly server can be run on a different machine and hoverfly binary won't be able to find files relative to simulation file that was uploaded from a client machine. Am I missing something here?

you're right, it won't work if hoverfly is remote. What about reading the bodyFile is done locally in `core/import.go` and `hoverctl/cmd/import.go`? I think that solves the first issue as well.

ns3777k:
I've got confused a little :smiley: 

So, first off is to patch hoverctl to loop through all pairs, replacing bodyFile with body filled with files' contents. Would be a little tricky 'cause we should unmarshal simulation data, but no problems so far.
Touches only client side, so no BC for server and no need for v6 and supporting bodyFile in schema, right?

Then patch hoverfly import cmd to do the same thing.

Question is: should I keep v6 with bodyFile?

tommysitu:
Not entirely sure if reading the `bodyFile` in the hoverctl import is the right way to go:
1. It won't work if someone imports the simulation using the Admin API
2. Any tools that build around the hoverfly Admin API (such as Hoverfly-Java) have to replicate the logic to handle `bodyFile`. 

sorry for contradicting to what I said 😳 . I underestimated the challenges of adding this feature.

So if we keep what you have done, then we must find a way to handle relative path. Any thoughts? 

Nevertheless, we should keep the v6 schema.

ns3777k:
> So if we keep what you have done, then we must find a way to handle relative path. Any thoughts?

Well. As I said before, what about hoverfly cwd process? That did work for me. When we call ioutil.ReadAll, it resolves the relative path against the cwd. Could you show me the failed case?

tommysitu:
Here's what I've tried: 
- make build
- start hoverfly binary in the target folder 
```
$ target/hoverfly 
```
- copy the body file to the same folder as the hoverfly binary: 
```
$ cp 200.json ~/go/src/github.com/SpectoLabs/hoverfly/target/200.json
```
- import the simulation that has `"bodyFile": "200.json",`, then it blows up:
```
$ hoverctl import time.json
Could not import simulation

{"error":"An error occurred: open 200.json: no such file or directory"}{"data":{"pairs":[],"globalActions":{"delays":[],"delaysLogNormal":[]}},"meta":{"schemaVersion":"v6","hoverflyVersion":"v1.1.3","timeExported":"2020-02-19T22:24:11Z"}}

```

ns3777k:
that's correct because your cwd is not `target`, but the directory where `target` is at. in other words valid cwd = $PWD.

tommysitu:
it's very odd, if I moved hoverfly to any other folder such as 
`~/Download/hoverfly`, then it can find the file in 
`~/Download/200.json`. 

but if hoverfly is in 
`~/go/src/github.com/SpectoLabs/hoverfly/target/hoverfly`, then the file needs to be in `~/go/src/github.com/SpectoLabs/hoverfly/200.json`

tommysitu:
ok, it's relative to where the hoverfly binary is executed.

ns3777k:
It can find the file in `Download` because you probably run the hoverfly from the same directory? If you're on linux (not sure about mac) you run the following command to know the wd of the process: pwdx `pgrep hoverfly`.

That's usually $PWD unless changed.

ns3777k:
race condition :smile: so, there are a couple things left here:

1. prevent deleting the simulation on failed import
2. write some tests
3. write some docs

right?

tommysitu:
pretty much. If you can fix #1 that will be great. 🌟 feel free to do 2 and 3 in a separate PR

ns3777k:
According to removal the simulation on a file error. I think I've found one easy way to prevent this. 

1. Methods inside `Hoverfly::DeleteSimulation` should return the removed elements and get saved in the new backup simulation which is returned from `Hoverfly::DeleteSimulation`.

2. New method `Hoverfly::RestoreSimulation` or `Hoverfly::ReplaceSimulation` should be added to replace the current simulation with the previous backed up in case of error.

Pseudo code:



`hoverfly_service.go`:

before:

```go
func (this *Hoverfly) DeleteSimulation() {
	this.Simulation.DeleteMatchingPairs()
	this.DeleteResponseDelays()
	this.DeleteResponseDelaysLogNormal()
	this.FlushCache()
}
```

after:

```go
func (this *Hoverfly) RestoreSimulation(s *models.Simulation) {
	this.Simulation.DeleteMatchingPairs()
	this.Simulation.ResponseDelays = s.ResponseDelays
	this.Simulation.ResponseDelaysLogNormal = s.ResponseDelaysLogNormal
	pairs := s.GetMatchingPairs()
	this.Simulation.AddPairs(pairs)
}

func (this *Hoverfly) DeleteSimulation() *models.Simulation {
	pairs := this.Simulation.DeleteMatchingPairs()
	delays := this.DeleteResponseDelays()
	delaysLogNormal := this.DeleteResponseDelaysLogNormal()
	this.FlushCache()

	s := models.NewSimulation()
	s.ResponseDelays = d1
	s.ResponseDelaysLogNormal = d2
	s.AddPairs(pairs)

	return s
}
```

In `simulation_handler.go` we save the simulation returned from `DeleteSimulation` and pass it to `RestoreSimulation` in case of errors.

There is a thing I didn't figure out for now. From what I see, ResponseDelays and ResponseDelaysLogNormal can be written in and read from but they are not protected with a mutex.

tommysitu:
Sorry for the late reply, I was under the weather. Just looked at your proposed solution, I'm a bit wary about the potential performance impact. We were aware that some hoverfly users have huge simulation files loaded in. It would be great if none of the existing hoverfly users gets much penalty for upgrading.

What about looping through the pair, and testing any presence of `bodyFile` to see if they are readable instead?

We already have `simulationView, err := NewSimulationViewFromRequestBody(body)` in the `addSimulation`, so using function such as  `os.OpenFile` to test `bodyFile`  should be relatively simple to do as well.

tommysitu:
After thinking about the relative path, we could make it a config option for user to specify where the root path it's relative to. That should solve the problem when user wants to store the file somewhere else rather than the cwd of hoverfly binary.

ns3777k:
1. This sounds much simpler but I'm not sure about performance impact, 'cause `os.OpenFile` and `f.Close` both make a syscall.

2. Agree. I'll make this option to be cwd by default.

tommysitu:
ignore os.openfile, that was my brain fart. What about os.stat()? It only reads the file metadata. I can’t tell what the actual performance is without a benchmark. I’m suggesting that we should do something to avoid impacting existing users when they upgrade. Alternatively It can do the backup only when body file.

ns3777k:
os.stat does not check whether a file is able to be read by the process (e.g. permissions) it can only tell that a file exists. i'll dig deeper soon :-)

tommysitu:
Thanks for looking into this!

ns3777k:
So, that's what I've come up with (latest commit):

1. We can avoid additional performance impact if we read the bodyFiles into body right after creating a simulation view (we'll read them anyway later). So we can early return on error before deleting simulations.

I moved the body reading part into `PutSimulation` method as well as deleting simulation call because I didn't want to read body files inside context of http handler. So basically I encapsulated this piece of logic there so everything is in 1 place.

2. I've added the configuration option `-response-body-files-path`, but I suck at naming things :smile: maybe you've gotta better one.

ns3777k:
what's the plan? :male_detective: should i start on the docs and tests? or you need some time to review this (last commit should solve previous issues)?

tommysitu:
I need to do some more testing around your last commit. How urgently do you need this feature?  I'm planning to a bug fix release for v1.1.5 this week. Then this one can go into the next one v1.2.0 with another two issues that need schema changes: 
https://github.com/SpectoLabs/hoverfly/issues/807 
https://github.com/SpectoLabs/hoverfly/issues/761

What do you think?

ns3777k:
No urgency. Just wanted to know the plan :smile:

core/hoverfly_service.go
```diff
@@ -283,7 +283,16 @@ func (hf Hoverfly) GetFilteredSimulation(urlPattern string) (v2.SimulationViewV6
 		hf.version), nil
 }
 
-func (this *Hoverfly) PutSimulation(simulationView v2.SimulationViewV6) v2.SimulationImportResult {
+func (this *Hoverfly) PutSimulation(simulationView v2.SimulationViewV6, overrideExisting bool) v2.SimulationImportResult {
```
tommysitu: This could break for someone using Hoverfly as a library. We should leave this public interface as it was, but created a new one for replacing simulation:
`Hoverfly.ReplaceSimulation(simulationView)`

Both funcs can call an internal one `func (this *Hoverfly) updateSimulation(simulationView v2.SimulationViewV6, overrideExisting bool) v2.SimulationImportResult ` to avoid duplicated code.

tommysitu:
Also I realized one more thing: you've loaded the file and put it in the `body` field. so it will show up in the exported simulation. Can we do something about this? Perhaps adding some way to keep track if the body was from the `bodyFile` and should be excluded on export.

core/hoverfly_service.go
ns3777k: true. fixed

core/cmd/hoverfly/main.go
```diff
@@ -179,9 +180,18 @@ func init() {
 func main() {
 	hoverfly := hv.NewHoverfly()
 
+	dir, err := os.Getwd()
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
```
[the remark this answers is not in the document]
ns3777k: Yes. But getting the cwd can return an error (e.g. https://manned.org/sys_getcwd/ef67ad78).

ns3777k:
Yes, that's exactly what I asked some days ago https://github.com/SpectoLabs/hoverfly/pull/893#issuecomment-586734806 :smile: I'll think what can be done to avoid this.

core/cmd/hoverfly/main.go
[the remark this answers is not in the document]
ns3777k: But I'll double check if we can ignore the error and keep the `dir` variable to be an empty string. Thus we just delay the error handling and it'll occur when a bodyFile is not able to be read from in case it contains a relative path.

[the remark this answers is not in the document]
tommysitu: what I meant is: 
```
ioutil.ReadFile(filepath.Join("", pair.Response.GetBodyFile()))
```
should read the file under the current working directory, right? 
so setting the `response-body-files-path` to "" should be equivalent to setting it to `os.Getwd()`?

tommysitu:
This PR is on the right track, I quite like to merge it to the next release branch (https://github.com/SpectoLabs/hoverfly/tree/v1.2.0). Do you mind reopen this issue for merging into the v1.2.0 branch? Then we can tackle the export issue.  Thanks!

tommysitu:
also it's quite a long thread, that's probably why I was losing track 😆

core/cmd/hoverfly/main.go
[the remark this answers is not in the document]
ns3777k: Oh, yes. I used the `os.Getwd` to show the current dir in `--help` information, otherwise there will be an empty string there.

ns3777k:
I switched the base branch to `v1.2.0

Issue #895, opened 2020-01-31
JSON request body whose root starts with an array fails to match with JsonPath/JsonPartial as matcher


cvadityaflipkart:
## Description of the bug
JSON request body whose root starts with an array fails to match with JsonPath as matcher.

## Steps to reproduce the issue
Any http request with the following body
https://jsonplaceholder.typicode.com/posts

and the simulation file containing the following matcher
```
        "body" : [ {
          "matcher" : "JsonPath",
          "value" : "$[0]"
        } ]
```
Edit: It also occurs with JsonPartial
```
        "body" : [ {
          "matcher" : "jsonpartial",
          "value" : "{\"userId\": 1}"
        } ]
```
### Observed result
**Hoverfly error messages seen (If none, say none)**
```
Failed to unmarshal body to JSON: json: cannot unmarshal array into Go value of type map[string]interface {}
```
**If possible, add screenshots to help explain your problem**

### Expected result
No error

## Additional relevant information
1. Hoverfly version: 1.1.3
2. Schema Version: v5
3. Probable fix at https://github.com/SpectoLabs/hoverfly/blob/7c00820058912ad13d8060b2ab996ff5c573b934/core/matching/matchers/json_path_match.go#L38

tommysitu:
fixed in v1.1.4

PR #896, opened 2020-01-31
V1.1.3-hotfix1

cvadityaflipkart:
https://github.com/SpectoLabs/hoverfly/issues/895


core/matching/matchers/json_path_match.go
```diff
@@ -35,7 +35,7 @@ func JsonPathExecution(matchString, toMatch string) (string, error) {
 		return "", err
 	}
 
-	var data map[string]interface{}
+	var data interface{}
```
tommysitu: 👍

core/matching/matchers/json_partial_match.go
```diff
@@ -7,22 +7,43 @@ import (
...
+		}
+		nodes := getAllNodesFromArray(actual)
+		for _, node := range nodes {
+			if mapContainsMap(expected, node) {
+				return true
+			}
+		}
+	default:
+		var actual map[string]interface{}
+		err3 := json.Unmarshal([]byte(toMatch), &actual) // has type interface{}
```
tommysitu: The json is already unmarshaled to `toMatchType`, you can do a type conversion here:
```
actual, ok := toMatchType.(map[string]interface{})
if !ok {
        // log an error
	return false
}
```

cvadityaflipkart: Done

cvadityaflipkart:
@tommysitu Welcome!!! Will this be released as a hotfix for v1.1.3 ?

tommysitu:
Thanks! will check it out today. I'm planning to release it in  v1.1.4 this week.

core/matching/matchers/json_partial_match.go
```diff
@@ -7,24 +7,37 @@ import (
...
 	if err0 != nil || err1 != nil {
 		return false
 	}
 
-	nodes := getAllNodesFromMap(actual)
-	for _, node := range nodes {
-		if mapContainsMap(expected, node) {
-			return true
+	actual, ok := toMatchType.(map[string]interface{})
+	if !ok {
```
tommysitu: let's do ` if ok .... else ...` rather then the opposite

tommysitu: I will do that on my end

tommysitu:
Thanks for the fix! @cvadityaflipkart A release should be out sometime today

cvadityaflipkart:
@tommysitu Welcome

PR #905, opened 2020-03-01
disable colors

### LINKED ISSUE #890: Add a way to disable colors in the plaintext log
Hoverfly always prints ansi color codes when `-logs plaintext` is used. This can be annoying on systems that do not support ansi colors. In my case, I'm working with the Docker image and I see output like this:

~~~
2019-12-18T15:13:14.824+01:00 hoverfly> [36mINFO[0m[2019-12-18T14:13:14Z] Listen on specific interface                  [36mhost[0m=0.0.0.0
2019-12-18T15:13:14.824+01:00 hoverfly> [36mINFO[0m[2019-12-18T14:13:14Z] Using memory backend
~~~

It would be great if there was some way to disable color output for plaintext logs. Probably it could be enough to just not *force* colors here: https://github.com/SpectoLabs/hoverfly/blob/e523f31338156dfad2d187c4f7f527c41fff4ee0/core/cmd/hoverfly/main.go#L189

This is not super important. There's always the possibility to view logs with `less -R` or other ways to preserve the colors.

ns3777k:
Closes #890

### DIFF
--- a/core/cmd/hoverfly/main.go
+++ b/core/cmd/hoverfly/main.go
@@ -103,6 +103,7 @@ var (
 
 	logsFormat = flag.String("logs", "plaintext", "Specify format for logs, options are \"plaintext\" and \"json\"")
 	logsSize   = flag.Int("logs-size", 1000, "Set the amount of logs to be stored in memory")
+	noColors   = flag.Bool("no-colors", false, "Disable colors for logging")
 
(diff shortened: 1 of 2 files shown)


tommysitu:
I'm gonna change the flag to `-log-color` for clarity. I'm adding it to the v1.1.5 release, thanks!

PR #911, opened 2020-03-09
WIP: Export body files / body / all

ns3777k:
Since bodyFiles get loaded into `body` on import, `body` is returned filled on export along with `bodyFile`. When both fields provided a warning is printed and hoverfly will use only body.

One way to get around this behavior is to add some flag to `/api/v2/simulation` that defines how to export the simulation. I've added `bodyMode` that can configure this based on the value:

1. `body-only`: export only `body`, emptying all bodyFile fields (which is default)
2. `prefer-bodyfile`: export both fields, but set `body` to an empty string when `bodyFile` is not empty (maybe this should be the defaut?)
3. `all`: export both `body` and `bodyFile` as is (current behavior)

It's just a quick working example to show a concept. It builds but tests don't pass for now. Refactoring required. Don't merge :warning:

### DIFF
--- a/core/hoverfly_funcs.go
+++ b/core/hoverfly_funcs.go
@@ -1,7 +1,6 @@
 package hoverfly
 
 import (
-	"errors"
 	"fmt"
 	v2 "github.com/SpectoLabs/hoverfly/core/handlers/v2"
 	"github.com/aymerick/raymond"
@@ -10,7 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	herrors "github.com/SpectoLabs/hoverfly/core/errors"
+	"github.com/SpectoLabs/hoverfly/core/errors"
 	"github.com/SpectoLabs/hoverfly/core/matching"
 	"github.com/SpectoLabs/hoverfly/core/matching/matchers"
 	"github.com/SpectoLabs/hoverfly/core/models"
@@ -46,7 +45,7 @@ func (hf *Hoverfly) DoRequest(request *http.Request) (*http.Response, error) {
 }
 
 // GetResponse returns stored response from cache
-func (hf *Hoverfly) GetResponse(requestDetails models.RequestDetails) (*models.ResponseDetails, *herrors.HoverflyError) {
+func (hf *Hoverfly) GetResponse(requestDetails models.RequestDetails) (*models.ResponseDetails, *errors.HoverflyError) {
 
 	var response models.ResponseDetails
 	var cachedResponse *models.CachedResponse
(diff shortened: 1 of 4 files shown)


ns3777k:
@tommysitu how do you think? would this approach that solve the exporting issue?

tommysitu:
Hey @ns3777k thank you for suggesting a solution. How about doing something simple first: 

if the response has both `body` and `bodyFile`, exporting should create or update the body file with the content from `body`, and then clearing the `body` field. This is useful if the user has run some capture, and doing an export can keep body files up-to-date 

This can be expanded to support another feature is to allow users doing capture, and exporting a simulation with body stores in separate files.

So importing/ exporting simulation with body files workflow will be complete.

ns3777k:
@tommysitu Sounds cool. As far as I understand, `hoverctl export` should create / update bodyFiles locally, right? What if bodyFile is an absolute path, where this file should be saved?

tommysitu:
yeah, that's what hoverctl does, it calls the get simulation API, and save it to a file.

ns3777k:
Yes, but if we run hoverctl export against a remote instance and get back an absolute path in bodyFile, we should not save it locally in the given absolute path for obvious security reasons, so the question is how to handle absolute paths when exporting?)

tommysitu:
Agreed. I feel a bit wary about allow hoverctl to write to any absolute path autonomously on export.  We should enforce `bodyFile` path to be relative only. What do you think?

ns3777k:
Actually that makes sense since we allow to set the path against the relative bodyFiles get resolved.

ns3777k:
I've force pushed the branch. some tests still fail, I'll make em work later. Could you take a look please?

tommysitu:
sure @ns3777k, will get back to you soon.

hoverctl/wrapper/simulation.go
```diff
@@ -38,14 +37,27 @@ func ExportSimulation(target configuration.Target, urlPattern string) ([]byte, e
...
 
-	var jsonBytes bytes.Buffer
-	err = json.Indent(&jsonBytes, body, "", "\t")
-	if err != nil {
+	var view v2.SimulationViewV6
+	if err := json.Unmarshal(body, &view); err != nil {
 		log.Debug(err.Error())
-		return nil, errors.New("Could not export from Hoverfly")
+		return nil, err
+	}
```
tommysitu: you can read response body into a struct in one go: 
```
err := json.NewDecoder(response.Body).Decode(&view)
```

ns3777k: true

core/hoverfly_funcs.go
```diff
@@ -127,7 +129,17 @@ func (hf *Hoverfly) readResponseBodyFiles(pairs []v2.RequestMatcherResponsePairV
 		}
 
 		if len(pair.Response.GetBody()) == 0 && len(pair.Response.GetBodyFile()) > 0 {
-			fileContents, err := ioutil.ReadFile(filepath.Join(hf.Cfg.ResponsesBodyFilesPath, pair.Response.GetBodyFile()))
+			bodyFile := pair.Response.GetBodyFile()
+			if filepath.IsAbs(bodyFile) {
+				err := errors.New(fmt.Sprintf(
```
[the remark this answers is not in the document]
ns3777k: true

ns3777k:
I've fixed the current tests and the code your comments were referenced to. (I didn't write new tests and docs. I'll make a separate PR)

PR #917, opened 2020-03-30
Hoverctl bodyfiles tests

ns3777k:
I had to write a fsHelper struct with an interface to replace it with a stub in tests.

### DIFF
--- a/hoverctl/cmd/export.go
+++ b/hoverctl/cmd/export.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/SpectoLabs/hoverfly/hoverctl/configuration"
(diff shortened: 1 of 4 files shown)


ns3777k:
conflicts fixed

hoverctl/cmd/export.go
```diff
@@ -22,10 +22,11 @@ will be written to the file path provided.
 
 		checkArgAndExit(args, "You have not provided a path to simulation", "export")
 
-		simulationData, err := wrapper.ExportSimulation(*target, urlPattern)
+		fsHelper := &configuration.LocalFileHelper{}
+		simulationData, err := wrapper.ExportSimulation(*target, fsHelper, urlPattern)
```
tommysitu, line 26: I was wondering why you need to mock the file io, and realized you are writing the body file in `wrapper.ExportSimulation` which forces you to unit test it. 

Somehow it was decided that we called the HTTP client for hoverfly API, a "wrapper" in this project. It's wrapping the API resource, so file writing probably not belong to it, and should be outside. Then we can add some functional tests for the export command without any mocking.

ns3777k: Oh. I got it.

ns3777k:
I moved the file writing out of `wrapper.ExportSimulation` function, fixed the unit tests and added a couple of functional ones to test the file writtings.

tommysitu:
I'm going to do another release before v1.2. It will upgrade to use go v1.4 and go module. Are you still up for https://github.com/SpectoLabs/hoverfly/issues/807? 

I got a branch for this: https://github.com/SpectoLabs/hoverfly/issues/761. Just need to tidy it up.

ns3777k:
I'm on final tests for hoverfly on bodyFile. Then I'll do #807 (I have actually done half of it). I hope I'll get everything done next week.

tommysitu:
awesome! 👍

PR #923, opened 2020-04-16
implementing fixed delay for specific response

ns3777k:
Only fixed delay for responses for now (I'll do logNormal after), no new tests and docs, I'll add them after review.
#807

### DIFF
--- a/core/modes/modes.go
+++ b/core/modes/modes.go
@@ -34,7 +34,7 @@ const Spy = "spy"
 const Diff = "diff"
 
 type Mode interface {
-	Process(*http.Request, models.RequestDetails) (*http.Response, error)
+	Process(*http.Request, models.RequestDetails) (ProcessResult, error)
 	SetArguments(arguments ModeArguments)
 	View() v2.ModeView
 }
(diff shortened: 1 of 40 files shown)


core/modes/modes.go
```diff
@@ -77,7 +90,7 @@ func ReconstructRequest(pair models.RequestResponsePair) (*http.Request, error)
 }
 
 // ReconstructResponse changes original response with details provided in Constructor Payload.Response
-func ReconstructResponse(request *http.Request, pair models.RequestResponsePair) *http.Response {
+func ReconstructResponse(request *http.Request, pair models.RequestResponsePair) ProcessResult {
```
tommysitu, line 93: can we not change the interface for this function? it's for reconstructing the HTTP response. It doesn't do much about the delay. You can create the `ProcessResult` within the `process` function, it can reduce the amount of changes in this PR.

ns3777k: fixed

ns3777k:
I'm gonna add docs and tests this weekend

ns3777k:
@tommysitu done. i'll make logNormal in a separate PR.

tommysitu:
Great! I will have another look soon.

tommysitu:
Hey, just let you know that I've updated master to use go v1.14 and go modules,  which means you can check out and build hoverfly in any folder. I've also fixed the delay test for modify mode, because in modify mode, it doesn't do matching.

ns3777k:
ok ok thx :+1: i'll add log normal delay soon.

PR #925, opened 2020-04-21
Fix authentication in https case

pashaosipyants:
Not sure this fix is correct, didn't dive deep into the code. 
So this PR needs **very** attentive review. 
With this PR, mainly, I wanted to demonstrate the issue #926

### DIFF
--- a/core/proxy.go
+++ b/core/proxy.go
@@ -169,7 +169,6 @@ func unauthorizedError(request *http.Request, realm, message string) *http.Respo
 func proxyBasicAndBearer(proxy *goproxy.ProxyHttpServer, realm string, basicFunc func(user, passwd string) bool, bearerFunc func(token string) bool) {
 
 	proxy.OnRequest().Do(goproxy.FuncReqHandler(func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
-		// IN HTTPS CASE AUTH HEADERS ARE IN CONNECT REQUEST, BUT NOT IN REQUEST TO DESTINATION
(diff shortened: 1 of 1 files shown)


core/proxy.go
```diff
@@ -169,6 +169,9 @@ func unauthorizedError(request *http.Request, realm, message string) *http.Respo
 func proxyBasicAndBearer(proxy *goproxy.ProxyHttpServer, realm string, basicFunc func(user, passwd string) bool, bearerFunc func(token string) bool) {
 
 	proxy.OnRequest().Do(goproxy.FuncReqHandler(func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
+		if strings.HasSuffix(req.URL.Host, ":443") {
```
pashaosipyants, line 172: in https case auth headers are in connect request, but not in request to destination

core/proxy.go
```diff
@@ -60,6 +52,14 @@ func NewProxy(hoverfly *Hoverfly) *goproxy.ProxyHttpServer {
 		})
 	}
 
+	proxy.OnRequest(matchesFilter(hoverfly.Cfg.Destination)).
```
pashaosipyants, line 55: authentication connect handler was unreachable, only this one was called

core/proxy.go
tommysitu: that's right, it's not available after the proxy CONNECT

core/proxy.go
tommysitu: The handlers are invoked in order it's registered to the proxy, this should fix it. It makes sense to have the auth handler intercept requests first.

PR #932, opened 2020-05-05
response logNormalDelay

ns3777k:
No new tests for now. I'll add them after review.
I've got a concern. I moved min, max, mean and median into separate struct into core/delay package and put it inside every struct where those fields were used. Maybe it'd be better to use this struct only where json representation is required and keep fields for internal structs as used to be.


core/models/delay_log_normal.go
```diff
@@ -14,11 +15,8 @@ import (
 type ResponseDelayLogNormal struct {
 	UrlPattern     string         `json:"urlPattern"`
 	HttpMethod     string         `json:"httpMethod"`
-	Min            int            `json:"min"`
```
tommysitu, line 17: lets keep internal struct as it is. It's easier to make it change in the future if these structs are not coupled with the simulation view.

core/hoverfly_service.go
```diff
@@ -213,19 +213,18 @@ func (hf *Hoverfly) SetResponseDelaysLogNormal(payloadView v1.ResponseDelayLogNo
...
-			Mean:       responseDelayView.Mean,
-			Median:     responseDelayView.Median,
-			DelayGenerator: delay.NewLogNormalGenerator(
-				responseDelayView.Min,
-				responseDelayView.Max,
-				responseDelayView.Mean,
-				responseDelayView.Median,
-			),
+			LogNormalDelayOptions: &logNormalOptions,
+			DelayGenerator: delay.NewLogNormalGenerator(logNormalOptions),
```
tommysitu, line 227: I think this is for the list of global delays, so maybe we should leave it as it is.

ns3777k:
I think I've fixed this. I had to add `HasLogNormalDelay` to `Response` interface because `ResponseDelay` cannot be checked for nil later.

core/handlers/v2/simulation_views_v6.go
```diff
@@ -87,3 +111,11 @@ func (this ResponseDetailsViewV6) GetHeaders() map[string][]string { return this
 
 // Gets FixedDelay - required for interfaces.Response
 func (this ResponseDetailsViewV6) GetFixedDelay() int { return this.FixedDelay }
+
+// Gets LogNormalDelay - required for interfaces.Response
+func (this ResponseDetailsViewV6) GetLogNormalDelay() interfaces.ResponseDelay {
+	return this.LogNormalDelay
```
tommysitu, line 117: a bit overkilled to add `HasLogNormalDelay` to the interface just for checking the delay is nil. You can try this, it returns an actual `nil` 
```
if this.LogNormalDelay != nil {
       return this.LogNormalDelay
}
return nil
```

tommysitu:
other than the nil checking bit, every else looks good 👍

core/handlers/v2/simulation_views_v6.go
ns3777k: wow, nice trick :smile:

ns3777k:
done, thanks :+1: . i'll add tests this week.

PR #935, opened 2020-05-15
fix logNormal response delay in middleware and add tests

ns3777k:


tommysitu:
Thank you for your contribution massively! I'm still trying to get everything in line to release this. Will keep you updated 🙂

PR #960, opened 2021-01-24
remove unnecessary binary

ns3777k:
I came across the 15Mb binary in the repo from 2018 :-) I guess this can be deleted?


tommysitu:
yes, it's supposed to be ignored: 
https://github.com/SpectoLabs/hoverfly/blob/master/.gitignore#L51
but that line ^^ doesn't match the name

PR #961, opened 2021-01-24
Body files remote

### LINKED ISSUE #951: Allow for remote bodyFiles
I want to be able to run hoverfly locally, but load remote simulations. This is possible using `--import` but bodyFiles cannot be loaded remotely. I want to do this so that third-party contributors can work off of the same set of simulations.

Just like with `--import`, I would like to be able to reference remote bodyFiles.

**Describe alternatives you've considered**
One alternative would be to write a tool to load files and inject JSON strings into simulation files and load those remotely.

ns3777k:
Closes #951

### DIFF
--- a/core/cmd/hoverfly/main.go
+++ b/core/cmd/hoverfly/main.go
@@ -58,6 +58,7 @@ var importFlags arrayFlags
 var destinationFlags arrayFlags
 var logOutputFlags arrayFlags
 var responseBodyFilesPath string
+var responseBodyFilesAllowedOriginFlags arrayFlags
 
 const boltBackend = "boltdb"
 const inmemoryBackend = "memory"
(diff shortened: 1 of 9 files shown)


tommysitu:
The changes look good to me. The only concern is that this feature might introduce a security issue if someone gains access to an external facing hoverfly instance, and they can use remote body file to fetch information from internal network. Convenience comes with risks. Any suggestion? Perhaps make it disabled unless you pass a start-up flag to switch on?

ns3777k:
hmmm. interesting catch 🧐 it can be a startup flag, or a little complex solution to have a whitelist for base urls (for protocol checking) / hostnames?

example for base urls:

```shell
$ hoverfly --allow-remote-body-file-origin=https://raw.githubusercontent.com:1234
```

or hostnames:

```shell
$ hoverfly --allow-remote-body-file-origin=raw.githubusercontent.com:1234
```

@tommysitu

tommysitu:
yeah that's a good idea 💡  I will go for matching both schema and hostname like CORS: `$ hoverfly --allow-remote-body-file-origin=https://raw.githubusercontent.com:1234`. To use this feature, user needs to pass this setting to avoid accidentally turning it on, and the file can only be fetched from the allow origin.

ns3777k:
@tommysitu done

tommysitu (approved):
It looks great! 👍

PR #974, opened 2021-08-12
fix random seed

### LINKED ISSUE #970: randomIntegerRange returns same value for requests less than 1 second apart
## Description of the bug
When multiple requests are made in quick succession to a URL that uses randomIntRange the same random value is returned in all responses.

This occurs even when response caching has been disabled with "-disable-cache"

The file core/util/random.go is re-initialising the seed of the RNG on every function call using the current unix epoch time in seconds.  This means multiple calls to the function within one second will return the same integer as the RNG gets re-seeded to the same value on each call.

Many other functions in core/util/random.go follow this same pattern of re-initialising the seed and will return the same value when called in quick succession.

Removing the line "rand.Seed(time.Now().Unix())" from the RandomIntegerRange function fixes the issue.

Launch hoverfly with a simulation that uses randomIntRange.
Contents of file randomIntRangeTest.json
```
{
  "data":{
    "pairs":[
      {
        "request":{
          "path":[{"matcher":"glob", "value":"/randomIntRangeTest"}],
          "method":[{"matcher":"exact", "value":"GET"}]
        },
        "response":{
          "status":200,
          "body":"{\"randomIntValue\":\"{{ randomIntegerRange 1000 9999 }}\"}",
          "encodedBody":false,
          "headers":{"Content-Type":["application/json"]},
          "templated":true
        }
      }
    ]
  }
}
```

```
./target/hoverfly -webserver -disable-cache -import randomIntRangeTest.json
```

Perform the request twice in quick succession.  EG:
```
curl -X GET 'http://localhost:8500/randomIntRangeTest'; curl -X GET 'http://localhost:8500/randomIntRangeTest'
```

### Observed result
The same randomInteger is returned for both requests
```
{"randomIntValue":"4362"}{"randomIntValue":"4362"}
```

### Expected result
Different values of randomIntValue should be returned for each request.

1. Hoverfly version: 1.3.2
2. Response caching has been disabled with "-disable-cache"

ns3777k:
Fixes #970

### DIFF
--- a/core/util/random.go
+++ b/core/util/random.go
@@ -15,7 +15,7 @@ const (
 )
 
 func RandomString() string {
-	rand.Seed(time.Now().Unix())
+	rand.Seed(time.Now().UnixNano())
 	return RandomStringWithLength(RandomIntegerRange(3, 16))
 }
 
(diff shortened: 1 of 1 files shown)


tommysitu:
thanks! @ns3777k

PR #983, opened 2021-09-10
State remove cors

ns3777k:
State removal does not return cors headers


ns3777k:
oops. gonna rebase it

PR #987, opened 2021-09-17
fix superfluous response.WriteHeader call

ns3777k:
Since `handlers.WriteResponse(w, marshal)` writes the response, additional `w.WriteHeader(http.StatusOK)` raises `superfluous response.WriteHeader call` error.


tommysitu (approved):
remove redundant `w.WriteHeader(http.StatusOK)` (according to go docs: 
```
        // Write writes the data to the connection as part of an HTTP reply.
	//
	// If WriteHeader has not yet been called, Write calls
	// WriteHeader(http.StatusOK) before writing the data.
```
LGTM

PR #988, opened 2021-09-18
fix pac file content-type

ns3777k:
response with pac file was supposed to have `application/x-ns-proxy-autoconfig` content-type, but it always was `text/plain`.

### DIFF
--- a/core/handlers/v2/hoverfly_pac_handler.go
+++ b/core/handlers/v2/hoverfly_pac_handler.go
@@ -43,8 +43,8 @@ func (this *HoverflyPACHandler) Get(w http.ResponseWriter, req *http.Request, ne
 	if pacFile == nil {
 		handlers.WriteErrorResponse(w, "Not found", 404)
 	}
-	handlers.WriteResponse(w, pacFile)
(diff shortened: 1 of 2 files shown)


core/handlers/v2/hoverfly_pac_handler.go
```diff
@@ -43,8 +43,8 @@ func (this *HoverflyPACHandler) Get(w http.ResponseWriter, req *http.Request, ne
 	if pacFile == nil {
 		handlers.WriteErrorResponse(w, "Not found", 404)
 	}
-	handlers.WriteResponse(w, pacFile)
-	w.Header().Set("Content-Type", "application/x-ns-proxy-autoconfig")
```
tommysitu, line 47: I thought this line should have set the correct type?

ns3777k: you cannot set headers after response is already written. it's written in `handlers.WriteResponse`

tommysitu: makes sense 👍

tommysitu (approved):

PR #1019, opened 2022-08-06
ISSUE-1014: Add docker multi-arch support 

ezienecker:
Add docker multi-arch support and add arm64 as target architecture

In addition to ticket 123, multi-arch support has also been added to the Docker image.

We use Hoverfly for our integration test where we spin up a Docker container from Hoverfly. To also support the colleagues with M1 we need a Docker image with arm64 architecture.

I tested the feature and execute the docker builds build command and uploaded the result to my personal docker hub repository. Please check this to verify that the implementation works: https://hub.docker.com/r/ezienecker/hoverfly/tags

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,8 +47,8 @@ jobs:
       - run:
           name: Build and push master Docker image
           command: |
-            docker build -t spectolabs/hoverfly:master .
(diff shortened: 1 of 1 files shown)


tommysitu (approved):
tested your image, this is perfect

PR #1024, opened 2022-09-10
feat(hoverfly): Introduce faker to templates

petkostas:
This PR extends the templating capabilities by introducing Faker.
Please mind that I am not an experienced GO developer (as my primary language is Python) but I tested the specific the addition and it works as expected:

```json
"response": {"status": 200, "body": "{\n  \"name\": \"{{faker 'Person' 'Surname'}}\", \"phone\": \"{{faker 'Phone' 'Number'}}\"\n}",
```

```
< HTTP/1.1 200 OK
< Access-Control-Allow-Origin: *
< Connection: keep-alive
< Content-Length: 55
< Content-Type: application/json
< Date: Sat, 10 Sep 2022 10:57:45 GMT
< Hoverfly: Was-Here
< Vary: Accept-Encoding
<
{
  "name": "Ozella Crist", "phone": "(900) 386-1612"
}
```


tommysitu:
this is a great idea. Have you looked into https://github.com/brianvoe/gofakeit#custom-functions as well? It seems to be more popular among the go community.

tommysitu:
If we use gofakeit, the template function can be easier to write as well, like 
```
{{ 'gofakeit' 'Email' }}

{{ 'gofakeit' 'Phone' }} 

```

what do you think?

petkostas:
@tommysitu Sounds really good, didn't see this one (and searched for go faker) I really prefer that, I will try to update the PR with that implementation! Thanks!

tommysitu (approved):
👍

tommysitu:
thanks @petkostas, I'm gonna change the param name from `faker` to `gofakeit` to reflect the implementation, and add some unit tests as well.

tommysitu:
actually I will just keep the name `faker` and just add some tests

Issue #903, opened 2020-02-20
application/x-www-form-urlencoded matcher


qcastel:
**Is your feature request related to a problem? Please describe.**
One of our request contains a content type 'application/x-www-form-urlencoded'.

Example of payload:

```
{
	"matcher": "exact",
	"value": "client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer&code=FAKE_CODE&grant_type=authorization_code&redirect_uri=https://local-auth.yapily.com:8081/&client_assertion=eyJraWQiOiJlcGwyT09Mem82WHlCWExDeFR4dUZ6SnJEQ0kiLCJhbGciOiJQUzI1NiJ9.eyJzdWIiOiIzZmMzMzljYS02MzM4LTRmNDMtOWUxYy1iYWZjNDM2MjA0MWMiLCJhdWQiOiJodHRwczpcL1wvbG9jYWwtZmFrZWJhbmsueWFwaWx5LmNvbTo4MDgxXC9vYXV0aDIiLCJpc3MiOiIzZmMzMzljYS02MzM4LTRmNDMtOWUxYy1iYWZjNDM2MjA0MWMiLCJleHAiOjE1ODIyMTY1ODMsImlhdCI6MTU4MjIxNjI4MywianRpIjoiYWQ3ZGI3NjctOTgxNC00ZDUzLWFmYTEtZjFlMDgxZjNmMDhkIn0.kVScfRmaQ3XX8JruabNg__baXo9pSgYHhmlgWHuxZwxd7RsFxIP-uCaGkzAIX8ZrLZ059VeeIZrGIySFYDnaTozb79MSN_Y9laDfEqj017e7zrfyuy1z21vpKyxgkaCAC1o226mIS9onldBRCHj3CXb-XL3arCM2X2fteNRURdrnBD9gJ_F3eXdyzTxbsUEe9siC8JCZ2YBXcp0SPODVooFA3NMvofSHRJf77a_0pQRAjFRX-xbQUYJGrNtXqfgiE0ImwkAStfZ-FTbGP0fMDE2t2BVSLat5pOBurhpE5_Hr8sPE0xb9YD4Uiyrco-pRwkMu8C95HitoOIiE2TXZqg"
}
```
Hoverfly seems to currently do a basic comparaison without parsing the payload.

**Describe the solution you'd like**

It would be nice to tell hoverfly which matching rule to apply for which parameter.
[...]

PR #1025, opened 2022-09-20
chore(hoverfly): add form processing for requests

petkostas:
Adds support for form processing in Hoverfly.
I recently encountered a case where I needed to parse form data for one of the virtual APIs. This became painful and required me to introduce a middleware.
Using the `ParseForm` of GO we can easily extend the request details with this logic.

⚠️ This is a draft, possibly there are other places we could add functionality (e.g. matchers?) ⚠️ 

To be added:

- Tests
- Extend request matchers (?)


tommysitu:
@petkostas I have fixed the tests already, you can try to rebase on the latest master. 

Because form present in the request body, I think we should just create a new form matcher for the body like this: 
```
          "body": [
            {
              "matcher": "form",
              "value": {
                "foo": [
                  {
                    "value": "bar",
                    "matcher": "exact"
                  }
                ],
                "low": [
                  {
                    "value": "bar",
                    "matcher": "exact"
                  }
                ]
              }
            }
          ],
```
what do you think?

tommysitu:
related to https://github.com/SpectoLabs/hoverfly/issues/903

petkostas:
@tommysitu Haven't abandoned this, I will most probably revisit this next week in order to push the changes.

petkostas:
@tommysitu picked this up again, will modify the matcher to the proposal (I initially performed a separate one, but your suggestion makes more sense).

petkostas:
@tommysitu I have the first draft, though as I am still fresh with GO would be best if you can check and recommend maybe a different way, I had a horrible time (and a lovely GO enlightening experience for a Python developer) finding a proper way to transform the nested Matchers.

petkostas:
Found a small issue with the caching (as it does not include the form data) will make a change to that accordingly as well

tommysitu:
Seems like a simple elegant solution, all existing tests passed 👍  I will poke around a bit more, hopefully nothing break. I will let you know.

core/models/payload.go
```diff
@@ -61,11 +61,13 @@ type RequestDetails struct {
 	Scheme      string
 	Query       map[string][]string
 	Body        string
+	FormData    map[string][]string
 	Headers     map[string][]string
 	rawQuery    string
 }
 
 func NewRequestDetailsFromHttpRequest(req *http.Request) (RequestDetails, error) {
+	req.ParseForm()
```
tommysitu, line 70: I found that this will break capture mode. Since it has read the body into a PostForm, 
L75: 
```
reqBody, err := util.GetRequestBody(req)
```
will return an empty body, and the captured simulation will be missing the request payload. 
You can try with this curl command: 
```
curl -X POST https://reqbin.com/echo/post/form --header 'Content-Type: application/x-www-form-urlencoded' --data 'param1=value1&param2=value2' --proxy localhost:8500
```

tommysitu: We save the captured simulation using this function: https://github.com/SpectoLabs/hoverfly/blob/master/core/hoverfly_funcs.go#L291

That need to be updated to save the form data in the body field.  So that when you switch back to simulate mode, the same request will get matched.

petkostas: Fixed in the latest commit! 😄

petkostas: Ok manual testing it in webserver mode raises an error after the capture, will dive into that as well

petkostas:
TODO:
- [x] Fix capture mode
- [ ] Update documentation

tommysitu (approved):
capture and simulate form data perfectly 👍  It would be good to extract the form matching logic into a `Form` matcher implementing the `MatcherFunc` type as well

Issue #842, opened 2019-05-31
Multiple entries of the same query parameter are encoded using ; (semicolon)


perlun:
## Description of the bug

In commit 1276c03cf5315658bc14c87c504b68d58ba03258 (PR #794), we fixed #792. However, the fix is still not optimal since query parameters where either part of the `key=value` part includes semicolon will be unable to be represented in the simulation JSON.

To be honest, my use case don't have a _requirement_ to use semicolons at the moment. However, I see a risk with the current implementation in that it uses a separator which could potentially be a fully valid part of a query parameter for a "reserved" purpose like this. This could potentially be very confusing to our users.

Ideally, the values would be present multiple times in the JSON (i.e. no separator), but if this is considered too complex/time consuming, we should at the very least choose a separator which is more than a single character. `;;;` would be an improvement (note, this is just a suggestion; feel free to come up with a better option if you like)

## Steps to reproduce the issue

Make a query which includes the same query string multiple times.
Like this: http://localhost/some-api/some-resource?q=access:vod&q=order:latest&q=profile:vd

### Observed result

```json
"q": [
    {
        "matcher": "exact",
        "value": "access:vod;order:latest;profile:vd"
    }
```

**Hoverfly error messages seen (If none, say none)**

```
none
```
**If possible, add screenshots to help explain your problem**

### Expected result

```json
"q": [
    {
        "matcher": "exact",
        "value": "access:vod;;;order:latest;;;profile:vd"
    }
```

or (preferably):

```json
"q": [
    {
        "matcher": "exact",
        "value": "access:vod"
    },
    {
        "matcher": "exact",
        "value": "order:latest"
    },
    {
        "matcher": "exact",
        "value": "profile:vd"
    }
```

or even this (I think this would be my personal favorite, in terms of readability of the JSON):

```json
"q": [
    {
        "matcher": "exact",
        "value": [
            "access:vod",
            "order:latest",
            "profile:vd"
        ]
    }
```

## Additional relevant information
1. Hoverfly version: 

```
+----------+--------------+
| hoverctl | master-3141  |
| hoverfly | v1.0.0-rc.2  |
+----------+--------------+
```

2. Anything that might help us to diagnose the problem

tommysitu:
Hi @perlun, joining multiple values using a symbol for a query is not a permanent solution, we prefer to implement what you suggest as well: 
```
"q": [
    {
        "matcher": "exact",
        "value": [
            "access:vod",
            "order:latest",
[...]

PR #1029, opened 2022-12-23
Added feature for matching query params with multiple values with new matchers

kapishmalik:
Added features for matching multiple with multiple values with newly introduced matchers containsexactly, containsonly, contains.

containsexactly - Verifies that the actual group contains exactly the given values and nothing else, in order.

containsonly - Verifies that the actual group contains only the given values and nothing else, in any order and ignoring duplicates (i.e. once a value is found, its duplicates are also considered found).

contains - Verifies that the actual group contains the given values, in any order.

All of the above matchers are case-sensitive. On saving simulation, we are saving as containsexactly as default value.

Implemented as described in the comment - https://github.com/SpectoLabs/hoverfly/issues/842#issuecomment-497652641


kapishmalik:
Resolves #842

tommysitu (approved):
Top-notch! love the tests 👍

tommysitu:
btw @kapishmalik you haven't set up your git email config properly, we can't really see you as a contributor for this commit.

kapishmalik:
@tommysitu is there a way now to correct this?

tommysitu:
looks like you have fixed it

kapishmalik:
I have fixed while raising other PR but not for this commit.

PR #1031, opened 2022-12-29
Handle array of string and empty interfaces for match

kapishmalik:
@tommysitu  I found one issue wherein on capturing the requests, we are saving an array of strings as values for matcher containsexactly as we are picking these values from query params, and on updating simulation via API call, an array of empty interfaces(generics) is being passed due to which code is breaking. Could you please take a look and approve this PR?

### DIFF
--- a/core/matching/matchers/contains_match.go
+++ b/core/matching/matchers/contains_match.go
@@ -1,6 +1,7 @@
 package matchers
 
 import (
+	"reflect"
 	"strings"
 
 	"github.com/SpectoLabs/hoverfly/core/util"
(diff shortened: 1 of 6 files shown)


kapishmalik:
@tommysitu it will be great if you can review this change asap as it is fixing breaking change.

core/matching/matchers/contains_match.go
```diff
@@ -9,10 +10,20 @@ import (
...
+	var matchStringArr []string
+	for i := 0; i < val.Len(); i++ {
+		currentValue := val.Index(i)
+		if currentValue.Kind() == reflect.Interface {
+			matchStringArr = append(matchStringArr, currentValue.Elem().String())
+		} else {
+			matchStringArr = append(matchStringArr, currentValue.String())
+		}
+
+	}
```
tommysitu, line 26: We should probably normalize these matchers into one so there is less duplication, also provide flexibility for extension, something like: 
```
"q": [
    {
        "matcher": "array",
        "options": {
          "ignoreOrder": true,
          "ignoreUnknown": true,
          "ignoreDuplicate": true
        }
        "value": [
            "access:vod",
            "order:latest",
            "profile:vd"
        ]
    }
```
what do you think?

tommysitu (approved):
Fixed a bug for the new array matchers, LGTM

core/matching/matchers/contains_match.go
kapishmalik: Yea, that works.

kapishmalik: @tommysitu  ll it be fine to take these options in the configuration itself?

kapishmalik: @tommysitu  raised PR #1033  for the same. I feel that we should keep those matchers (containsexactly, containsonly, contains) as well and let users decide which one to use based on their ease.

Issue #900, opened 2020-02-17
Add support for custom variables


maxmoehl:
**Is your feature request related to a problem? Please describe.**
I have a simulation that sends back lists, these lists get filled with data extracted from the request using xpath (same statement over and over again), due to huge xml files these xpath statements are quite long and complicated. It would be really nice to be able to define some variables at the top that can later be used for templating.

**Describe the solution you'd like**
A new field in the request object (or somewhere else) with the possibility to load values from the request and store them in variables. For example:
```json
"request.variables": [
  {
    "name": "varOne",
    "method": "xpath",
    "path": "/some/xapth/expression"
  }
]
```
`method` could be xpath or jsonpath and path needs to be set accordingly
later in the template you can access those vars with: `{{ var "varOne" }}`

**Describe alternatives you've considered**
Copy paste those xpath statements but making changes is a nightmare.
I guess using middelware would also somehow work but it is fairly complicated since you don't have access to the request.

**Additions**
This could also be used for other data types like dates. To be able to do something like this it would need to be more generic which complicates things. Maybe something similiar to the templating syntax:
```
"request.variables": [
  {
    "name": "valueFromXpath",
    "method": "Request.Body",
    "args": ["xpath", "/some/xapth/expression"]
  },
  {
    "name": "dateValue",
    "method": "now",
    "args": ["", "2006-01-02T15:04:05Z"]
  },
]
```

kapishmalik:
@tommysitu  I can work on this feature request. It will be really helpful as we also have such use cases.

tommysitu:
@kapishmalik  are you considering to store the custom variables in the simulation file?

kapishmalik:
@tommysitu yea, I would be defining them in simulation files. I need to think more about exact format and things which we can support. You can assign this to me . I will get back with complete implementation details around the same.

kapishmalik:
@tommysitu  I am planning to add support for custom literals defined by the user and custom variables which can be values fetched from either XPath or JsonPath. We will be later adding support for global literals which can be used in any template response. Let me know if we are good to implement this.

Literals can be accessed as - {{literal "literal1" }}
Vars can be accessed as {{ var "varOne" }}

I will be adding support later to add support for custom methods as well.

```"literals": [
            {
              "name": "<literal1>",
              "value": ["value1", "value2", "value3"]
            },
            {
              "name": "<literal2>",
              "value": "value2"
            },
            {
              "name": "<literal3>",
              "value": {
[...]

PR #1032, opened 2022-12-30
feature - add support for adding custom literals and variables

kapishmalik:
Resolves => issue #900 

Added support for custom literals defined by the user and custom variables which can be values fetched from the request body using either XPath or JsonPath.  We need to define literals and variables in the data section in the below format in the simulation file.

Literals can be accessed in templates as {{Literals.<literal name>}}
Variables can be accessed in templates as {{Variables.<variable name>}}

``` 
"variables": [
            {
                "name" : "varOne",
                "function": "faker",
                "arguments": ["Name"]
            },
            {
                "name": "varTwo",
                "function": "randomIPv4",
                "arguments": []
            },
            {
                "name": "uid",
                "function": "body",
                "arguments": ["xpath", "/root/id"]
            }
        ],
        "literals": [
            {
                "name":"literal1",
                "value": "value1"
            },
            {
                "name": "literal2",
                "value": ["value2", "value3", "value4"]
            }

        ]

### DIFF
--- a/core/handlers/v2/simulation_views_v5.go
+++ b/core/handlers/v2/simulation_views_v5.go
@@ -40,6 +40,17 @@ type MatcherViewV5 struct {
 	Config  map[string]interface{} `json:"config,omitempty"`
 }
 
+type LiteralViewV5 struct {
+	Name  string      `json:"name"`
+	Value interface{} `json:"value"`
+}
+
+type VariableViewV5 struct {
+	Name       string `json:"name"`
+	Method     string `json:"method"`
+	Expression string `json:"expression"`
+}
+
 func NewMatcherView(matcher string, value interface{}) MatcherViewV5 {
 	return MatcherViewV5{
 		Matcher: matcher,
@@ -47,13 +58,13 @@ func NewMatcherView(matcher string, value interface{}) MatcherViewV5 {
 	}
 }
 
-//Gets Response - required for interfaces.RequestResponsePairView
+// Gets Response - required for interfaces.RequestResponsePairView
 func (this RequestMatcherResponsePairViewV5) GetResponse() interfaces.Response { return this.Response }
 
 type ResponseDetailsViewV5 struct {
 	Status           int                    `json:"status"`
 	Body             string                 `json:"body"`
-	BodyFile         string              	`json:"bodyFile,omitempty"`
+	BodyFile         string                 `json:"bodyFile,omitempty"`
 	EncodedBody      bool                   `json:"encodedBody"`
 	Headers          map[string][]string    `json:"headers,omitempty"`
 	Templated        bool                   `json:"templated"`
@@ -61,9 +72,11 @@ type ResponseDetailsViewV5 struct {
 	RemovesState     []string               `json:"removesState,omitempty"`
 	FixedDelay       int                    `json:"fixedDelay,omitempty"`
 	LogNormalDelay   *LogNormalDelayOptions `json:"logNormalDelay,omitempty"`
+	Literals         []LiteralViewV5        `json:"literals,omitempty"`
+	Variables        []VariableViewV5       `json:"variables,omitempty"`
 }
 
-//Gets Status - required for interfaces.Response
+// Gets Status - required for interfaces.Response
 func (this ResponseDetailsViewV5) GetStatus() int { return this.Status }
 
 // Gets Body - required for interfaces.Response
@@ -97,6 +110,51 @@ func (this ResponseDetailsViewV5) GetLogNormalDelay() interfaces.ResponseDelay {
 	}
 
 	return nil
+}
+
+func (this ResponseDetailsViewV5) GetLiterals() []interfaces.Literal {
+
+	if this.Literals != nil {
+		var literals []interfaces.Literal
+		for _, literal := range this.Literals {
+			literals = append(literals, literal)
+		}
+		return literals
+
+	}
+	return nil
+}
+
+func (this ResponseDetailsViewV5) GetVariables() []interfaces.Variable {
+
+	if this.Variables != nil {
+		var variables []interfaces.Variable
+		for _, variable := range this.Variables {
+			variables = append(variables, variable)
+		}
+		return variables
+	}
+	return nil
+}
+
+func (this LiteralViewV5) GetName() string {
+	return this.Name
+}
+
+func (this LiteralViewV5) GetValue() interface{} {
+	return this.Value
+}
+
+func (this VariableViewV5) GetName() string {
+	return this.Name
+}
+
+func (this VariableViewV5) GetMethod() string {
+	return this.Method
+}
+
+func (this VariableViewV5) GetExpression() string {
+	return this.Expression
 }
 
 type LogNormalDelayOptions struct {
--- a/core/handlers/v2/simulation_views_v5.go
+++ b/core/handlers/v2/simulation_views_v5.go
@@ -43,9 +43,9 @@ type MatcherViewV5 struct {
 }
 
 type GlobalVariableViewV5 struct {
-	Name      string   `json:"name"`
-	Function  string   `json:"function"`
-	Arguments []string `json:"arguments"`
+	Name      string        `json:"name"`
+	Function  string        `json:"function"`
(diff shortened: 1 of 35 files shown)

### CODE REVIEW

**Comment by kapishmalik** [comment]
Improvements/suggestions are welcome.

File: core/handlers/v2/simulation_views_v5.go [hunk 1]
```diff
@@ -40,30 +40,43 @@ type MatcherViewV5 struct {
...
+	BodyFile         string                 `json:"bodyFile,omitempty"`
 	EncodedBody      bool                   `json:"encodedBody"`
 	Headers          map[string][]string    `json:"headers,omitempty"`
 	Templated        bool                   `json:"templated"`
 	TransitionsState map[string]string      `json:"transitionsState,omitempty"`
 	RemovesState     []string               `json:"removesState,omitempty"`
 	FixedDelay       int                    `json:"fixedDelay,omitempty"`
 	LogNormalDelay   *LogNormalDelayOptions `json:"logNormalDelay,omitempty"`
+	Literals         []LiteralViewV5        `json:"literals,omitempty"`
+	Variables        []VariableViewV5       `json:"variables,omitempty"`
```
tommysitu, line 76: Do we want to apply this for a particular response only?  I thought the idea of having a custom variable is that we can declare a template function globally and use it across multiple responses. This will give you some benefits in terms of managing templating data for a large simulation file, like you don't need to go through each response to modify the function when change is required.

For example, the `DataViewV5` could be changed to: 
```
type DataViewV5 struct {
	RequestResponsePairs []RequestMatcherResponsePairViewV5 `json:"pairs"`
	GlobalActions        GlobalActionsView                  `json:"globalActions"`
	GlobalVariables     GlobalVariablesView               `json:"globalVariables"`

}
```
where the `GlobalVariables` will be the root for the literals and the variables.

core/handlers/v2/simulation_views_v5.go
```diff
@@ -40,30 +40,43 @@ type MatcherViewV5 struct {
...
 
+type LiteralViewV5 struct {
+	Name  string      `json:"name"`
+	Value interface{} `json:"value"`
+}
+
+type VariableViewV5 struct {
+	Name       string `json:"name"`
+	Method     string `json:"method"`
+	Expression string `json:"expression"`
```
tommysitu, line 51: How about calling this `Function` and `Arguments`/`Args` instead? Because it could be used to declare any templating functions with arguments, not only for the xpath or jsonpath functions

core/handlers/v2/simulation_views_v5.go
kapishmalik: @tommysitu I was with assumption that he wants to have per request/response as mentioned about long response with long xpath/json path expression and he wants to replace multiple occurrences in a response. Yea, in that case we can have globally with func and args.

core/handlers/v2/simulation_views_v5.go
kapishmalik: We can have func and args globally as it will be much easier.

kapishmalik: @tommysitu  I have made the required changes. If the changes look good, then I can add more UTs and documentation.

kapishmalik: @tommysitu  You can review completely. Tested and added documentation for the same.

tommysitu (commented):
looks great! Could you update the doc to show how to use the `vars` helper method for templating as well? thanks!

tommysitu:
we also need to update the simulation schema definition and bump the simulation version to `v5.2`, I can do that once you are done.

core/templating/templating.go
```diff
@@ -38,48 +48,89 @@ func NewTemplator() *Templator {
...
+
+	variableMap := make(map[string]interface{})
+	if vars != nil {
+		for _, variable := range *vars {
+			if variable.Function != REQUEST_BODY_HELPER {
+				value, error := t.callHelper(variable)
+				if error != nil {
+					return error
+				}
+				variableMap[variable.Name] = value
```
tommysitu, line 104: this is going to execute the helper method once, and stored a constant value in the map. It's not gonna work if one is expecting random data or the current timestamp. I think we need to store the variables in the map and then execute the function only when we generate the context data for template rendering.

core/templating/templating.go
```diff
@@ -38,48 +48,89 @@ func NewTemplator() *Templator {
...
+	t.literals = literalMap
+
+}
+func (t *Templator) SetRequestIndependentVariables(vars *models.Variables) error {
+
+	variableMap := make(map[string]interface{})
+	if vars != nil {
+		for _, variable := range *vars {
+			if variable.Function != REQUEST_BODY_HELPER {
+				value, error := t.callHelper(variable)
```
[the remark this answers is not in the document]
kapishmalik: done

core/templating/templating.go
```diff
@@ -38,48 +48,89 @@ func NewTemplator() *Templator {
...
+		helperMethodMap["randomIntegerRange"] = t.randomIntegerRange
+		helperMethodMap["randomFloat"] = t.randomFloat
+		helperMethodMap["randomFloatRange"] = t.randomFloatRange
+		helperMethodMap["randomEmail"] = t.randomEmail
+		helperMethodMap["randomIPv4"] = t.randomIPv4
+		helperMethodMap["randomIPv6"] = t.randomIPv6
+		helperMethodMap["randomUuid"] = t.randomUuid
+		helperMethodMap["replace"] = t.replace
+		helperMethodMap["faker"] = t.faker
+		helperMethodMap["body"] = t.requestBody
```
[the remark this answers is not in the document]
kapishmalik: done

core/templating/templating.go
kapishmalik: done

kapishmalik: @tommysitu  incorporated all provided review comment changes

kapishmalik: @tommysitu  I have also handled concurrency while updating literals and variables. Added changes related to the same.

tommysitu: have been busy, i will review your PRs today @kapishmalik

kapishmalik: np... Take ur time.

tommysitu (approved):
Looks great. Thanks for taking concurrency into account as well. I did spot one more thing while I was playing with it, the `hoverctl export` command doesn't write the new fields into a simulation file. Any idea what it might be? Meanwhile, I think this PR is safe to merge. Well done man!

kapishmalik:
I will figure out why hoverctl export is not writing new fields in simulation file and will get back with the same.

kapishmalik:
@tommysitu  I checked this and found something weird. When I run the command by going into the target directory then it does not export variables whereas if I run target/hoverctl export <file name>... it is able to extract newly added fields as well.  
I also tried `go run main.go export <filename>` and it was able to get new fields. It is something to do with go modules and its caching. I m new to Go so could not identify the exact root cause but still reading on the same

kapishmalik:
> @tommysitu I checked this and found something. When I run the command by going into the target directory then it does not export new fields whereas if I run target/hoverctl export ... it is able to extract newly added fields as well as it may be able to reference the package with new changes(one containing Simulation view) from here. I also tried `go run main.go export <filename>` and it was able to get new fields. I m new to Go so could not identify the exact root cause but still reading on the same

@tommysitu can u help me fixing this. Facing exactly same issue with matcher chaining.

tommysitu:
yeah I need to sort this out before a release

kapishmalik:
> yeah I need to sort this out before a release

@tommysitu I feel this ll be fixed once we add a new release or tag.

kapishmalik:
> Looks great. Thanks for taking concurrency into account as well. I did spot one more thing while I was playing with it, the `hoverctl export` command doesn't write the new fields into a simulation file. Any idea what it might be? Meanwhile, I think this PR is safe to merge. Well done man!

@tommysitu I believe this might have happened because you might be using already installed hoverctl binary which be pointing to some other location. I faced this while working for some other PR and able to get this why this happened

tommysitu:
yeah you are right, it looks good with the master build

PR #1033, opened 2022-12-30
feature - introduce generic array matcher with options

kapishmalik:
Matches the matcher group with value passed in request array based on the configuration passed.

ignoreorder - ignore order in which values are passed
ignoreunknowns - ignore unknowns in the values passed
ignoreoccurrences - ignore comparision of no of occurrences in matcher group and values passed

``` 
   
   "matcher": "contains"
   "value": "[?]"
   "configuration": "{}"
{
    "matcher": "array",
    "config": {
        "ignoreunknown": <true/false>,
        "ignoreorder": <true/false>,
        "ignoreoccurrences": <true/false>
    }
    "value": [
        "access:vod",
        "order:latest",
        "profile:vd"
    ]
}

### DIFF
--- a/core/matching/matchers/matchers.go
+++ b/core/matching/matchers/matchers.go
@@ -2,6 +2,8 @@ package matchers
 
 type MatcherFunc func(data interface{}, toMatch string) bool
 
+type MatcherFuncWithConfig func(data interface{}, toMatch string, config map[string]interface{}) bool
+
 var Matchers = map[string]MatcherFunc{
 	// Default matcher
 	"": ExactMatch,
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: core/matching/matchers/matchers.go [hunk 1]
```diff
@@ -18,4 +20,9 @@ var Matchers = map[string]MatcherFunc{
 	ContainsExactly: ContainsExactlyMatch,
 	ContainsOnly:    ContainsOnlyMatch,
 	Contains:        ContainsMatch,
+	Array:           ContainsExactlyMatch,
```
tommysitu, line 23: Are you going to replace the other `Contains...` matchers with `Array`? so `Array` without config is the same as `ContainsExcactly` and if you want to do `Contains` then you pass `ignoreOrder: true, ignoreUnknown: true` as the config to the `Array` matcher.

tommysitu (approved):
really good start 👍

core/matching/matchers/matchers.go
kapishmalik: Yea, we will be deprecating it soon.

PR #1036, opened 2023-01-08
Code changes for matcher chaining

### LINKED ISSUE #761: Matcher Chaining

kapishmalik:
Resolves #761. Different way of implementation but following same approach as followed in PR #1035 

PR contains changes for implementing matcher chaining. Any matcher can be combined with any of the other matchers. Output/matched value of one matcher is feed into other matcher mentioned in do match. If this PR looks fine then I will add documentation. 

`{
body: [
{
matcher: "jsonpath"
value: “$.user.id"
domatch: {
matcher: “exact”,
value: “1"
}
]
}

"body": [
{
"matcher": "xpath",
"value": "/root/id",
"domatch": {
"matcher": "exact",
"value": "123"
}
}
]
`

### DIFF
--- a/core/matching/field_matcher_test.go
+++ b/core/matching/field_matcher_test.go
@@ -193,22 +193,6 @@ var fieldMatcherTests = []fieldMatcherTest{
 		toMatch: `<document><details>{"name":"Test", "id":"12345"}</details></document>`,
 		equals:  BeTrue(),
 	},
-	{
-		// exactmatch, containsexactly, jsonmatch, xmlmatch match exact details
-		name: "IgnoreDoMatchIncaseOfExactMatcher",
-		matchers: []models.RequestFieldMatchers{
-			{
-				Matcher: "exact",
-				Value:   "test",
-				DoMatch: &models.RequestFieldMatchers{
-					Matcher: "glob",
-					Value:   "*es*",
-				},
-			},
-		},
-		toMatch: "test",
-		equals:  BeTrue(),
-	},
 }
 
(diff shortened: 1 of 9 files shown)


kapishmalik:
@tommysitu I have added documentation and tested once. You can take a look.

core/matching/field_matcher_test.go
```diff
@@ -155,6 +155,44 @@ var fieldMatcherTests = []fieldMatcherTest{
...
+						DoMatch: &models.RequestFieldMatchers{
+							Matcher: "exact",
+							Value:   "Test",
+						},
+					},
+				},
+			},
+		},
+		toMatch: `<document><details>{"name":"Test", "id":"12345"}</details></document>`,
+		equals:  BeTrue(),
```
tommysitu, line 194: 🤩

tommysitu (approved):
Beautiful. Much better than the original implementation

core/matching/field_matcher_test.go
kapishmalik: ☺️

PR #1037, opened 2023-01-10
Make array matcher to be used by matcher chaining and upgrade schema to v5.2

kapishmalik:
@tommysitu  As of now, FieldMatcher takes the toMatch value as a semi-colon delimited string as one of the arguments. ArrayMatcher internally splits the string and again converts it back to an array. It was fine till now. 

But after introducing matcher chaining, there might be a scenario where the user first matches via JSON path and get value as JSON array and now wants to use array matcher to do further matching. To incorporate that scenario, I have made the necessary changes. I have passed json string array to FieldMatcher from QueryMatching as of now. As the probability of more than one value for same name is there for queries. Later, we can think of doing from other places as well. Also, none of other matchers apart from ArrayMatcher is splitting string and converting into array.

Also, as part of this PR, I upgraded simulation schema version to 5.2

### DIFF
--- a/core/matching/query_matching.go
+++ b/core/matching/query_matching.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"github.com/SpectoLabs/hoverfly/core/models"
+	"github.com/SpectoLabs/hoverfly/core/util"
 )
 
 func QueryMatching(requestMatcher models.RequestMatcher, toMatch map[string][]string) *FieldMatch {
@@ -45,7 +46,7 @@ func QueryMatching(requestMatcher models.RequestMatcher, toMatch map[string][]st
 			continue
 		}
 
-		fieldMatch := FieldMatcher(matcherQueryValue, strings.Join(toMatchQueryValues, ";"))
+		fieldMatch := FieldMatcher(matcherQueryValue, getStringToBePassedForMatching(toMatchQueryValues))
 		matcherHeaderValueMatched = fieldMatch.Matched
 		score += fieldMatch.Score
 
@@ -58,4 +59,16 @@ func QueryMatching(requestMatcher models.RequestMatcher, toMatch map[string][]st
 		Matched: matched,
 		Score:   score,
 	}
+}
+
+func getStringToBePassedForMatching(strArr []string) string {
+
+	if len(strArr) <= 1 {
+		return strings.Join(strArr, ";")
+	}
+	byteArr, err := util.JSONMarshal(strArr)
+	if err != nil {
+		return strings.Join(strArr, ";")
+	}
+	return string(byteArr[:])
 }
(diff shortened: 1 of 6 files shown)


kapishmalik:
@tommysitu can u just test once if all tests of field_matcher_test.go are working before merging PR? Ideally, it should be.

core/matching/query_matching.go
```diff
@@ -59,3 +60,15 @@ func QueryMatching(requestMatcher models.RequestMatcher, toMatch map[string][]st
 		Score:   score,
 	}
 }
+
+func getStringToBePassedForMatching(strArr []string) string {
```
tommysitu, line 64: I think converting query params or header values to json array back and forth is a bit overkill, it won't be as efficient as strings with a delimiter. It might not be worth making this change for a very specific use case, also the JSON array may contains JSON objects instead of simple string values. I feel like user using JSON path matchers would be better off chaining JSON paths to process the JSON data rather than mixing with the Array matcher.

tommysitu: btw you can take a look at this one for bumping minor version of the simulation: https://github.com/SpectoLabs/hoverfly/commit/d6c2ce4ca2e30a62545cb371ee80292a9bf1506d, we also need to update the simulation schema file: https://github.com/SpectoLabs/hoverfly/blob/master/schema.json

kapishmalik: > I think converting query params or header values to JSON array back and forth is a bit overkill, it won't be as efficient as strings with a delimiter. If the change is risky or has an adverse effect on wider hoverfly usages, I don't think it's worth making this change for a very specific use case.
> 
> In a real scenario, JSON array may contain JSON objects instead of just simple string values, so Array matcher might not be the right one to use anyway. If a user requires JSON path matching, they would be better off chaining the jsonPath matchers to process the JSON data rather than mixing with the Array matcher in my opinion.

@tommysitu yea completely makes sense. It will be an overkill and may have adverse affect as well so better if I wont do it.

kapishmalik: > btw you can take a look at this one for bumping minor version of the simulation: [d6c2ce4](https://github.com/SpectoLabs/hoverfly/commit/d6c2ce4ca2e30a62545cb371ee80292a9bf1506d), we also need to update the simulation schema file: https://github.com/SpectoLabs/hoverfly/blob/master/schema.json

@tommysitu it will be better if you can help me with this change. I don't want to mess things up while doing this.

kapishmalik: > btw you can take a look at this one for bumping minor version of the simulation: [d6c2ce4](https://github.com/SpectoLabs/hoverfly/commit/d6c2ce4ca2e30a62545cb371ee80292a9bf1506d), we also need to update the simulation schema file: https://github.com/SpectoLabs/hoverfly/blob/master/schema.json

I will make this change and and get back with new PR.

PR #1040, opened 2023-01-17
feature - Exclude headers and response fields from diff

### LINKED ISSUE #897: Exclude fields in Hoverfly Diff Mode
The diff mode in Hoverfly asserts on all the fields. It would be quite useful if there is a way to exclude fields during the assert.

A solution on the lines of matching strategies present in simulation mode could be used when in diff mode. Also support for JSONPath and XPath should be used as matchers in the exclusion.

kapishmalik:
This PR resolves #897 

@tommysitu I have raised PR to exclude headers and response fields from diff. 

The user will be making a POST call with the body to exclude diff entries. As of now, added support only for JSONPath as our diff properly does diff for JSON response. I didn't use a third-party library as of now to do JSON diff. Our code to do JsonDiff also looks fine but is unable to do diff for particular elements of the array etc. 

Also, I have added a provision to exclude headers from API as well.

Remaining: Adding UTs and documentation.

Curl command: curl --location --request POST 'http://localhost:8888/api/v2/diff' \
--header 'Content-Type: text/plain' \
--data-raw '{
    "excludedHeaders":["Date"],
    "excludedResponseFields":["$.time"]
}'


core/handlers/v2/diff_handler.go
```diff
@@ -59,6 +58,42 @@ func (this *DiffHandler) Delete(w http.ResponseWriter, req *http.Request, next h
...
+		Diff: diffsToReturn,
+	})
+	if err != nil {
+		handlers.WriteErrorResponse(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	handlers.WriteResponse(w, marshal)
+}
+
+func ConvertToResponseDiffView(responsesDiff map[SimpleRequestDefinitionView][]DiffReport) []ResponseDiffForRequestView {
```
tommysitu, line 85: should keep it unexported `convertToResponseDiffView`

tommysitu (commented):
I think it's a good experimental feature. Maybe that's all you need to do or maybe not. I don't currently have a use case for diff mode so can't really tell.

kapishmalik:
> I think it's a good experimental feature. Maybe that's all you need to do or maybe not. I don't currently have a use case for diff mode so can't really tell.

To be honest, same for me. I am also not 100% sure about correct use case for the same. Lets see if this works for one who created issue.

core/handlers/v2/diff_handler.go
kapishmalik: yea, done. Also added UTs and documentation.

tommysitu (approved):
👍

PR #1043, opened 2023-01-20
fix: update minify dependency

joobus:
The dependency `github.com/tdewolff/minify` currently used by hoverfly has several CVEs according to the code scanning tool Whitesource.  This PR updates that dependency.


tommysitu:
Thanks for spotting that!

PR #1048, opened 2023-02-09
fix: upgrade k8s/client-go

joobus:
The k8s/client-go library was flagged by Whitesource as having unpatched vulnerabilities.  This PR upgrades that package.  Dependencies required by the upgrade require Go 1.18 or higher so I bumped the version also.  Finally, I added the sandbox directory created when running `make test` to the `.gitignore` file.


tommysitu (approved):
Bump go dependencies, LGTM

PR #1051, opened 2023-02-12
Deprecate contains exactly matcher

### LINKED ISSUE #1050: Deprecate containsExactly matcher
The generic array matcher with default configs does the same thing as the `containsExactly` matcher which should be deprecated.

kapishmalik:
Resolves #1050 

Deprecate contains exactly matcher


tommysitu (approved):
👍

Issue #1044, opened 2023-01-22
Support for webhooks


petkostas:
**Is your feature request related to a problem? Please describe.**
Support for webhooks would benefit the use of some more complex API simulations, imagine an API that, after receiving a request, fires a callback/webhook.
In the current implementation, a way to tackle this would be using a Middleware, if we provide this feature built-in, we could eliminate the requirement for a Middleware for less complex scenarios (e.g. triggering a webhook with elements of the requesting payload).

**Describe the solution you'd like**
Extend the schema in order to support webhooks, this could be similar to how [wiremock supports them](https://wiremock.org/docs/webhooks-and-callbacks/) (check JSON part).

petkostas:
I started a POC for this

kapishmalik:
@petkostas hope you tested this out. If yes and it fullfill our requirements then we are good to close this issue.

petkostas:
> @petkostas hope you tested this out. If yes and it fullfill our requirements then we are good to close this issue.

@kapishmalik I did a small test with the latest changes, but I haven't tested the specific functionality after it got merged, but I believe we can close this and if needed a new issue can be opened. 👍

PR #1053, opened 2023-02-18
Add support for post serve action

kapishmalik:
Resolves => #1044 

Added below features

Expose GET/POST/DELETE APIs for fetching or setting flexible post-execution hooks (i.e. it'll be a script in the same way that we have for middleware).
Change in simulation schema to accept registered post-execution hook for particular response details.
Changes related to executing post-execution hook after execution is over.

Pending:
UTs fixes 
Documentation for newly added feature and APIs

@tommysitu  @petkostas  could u guys please review it and let me know if it looks fine?


kapishmalik:
@petkostas Can you do the testing once on your machine and see if it solves your use case?

Steps to test on your machine

- Clone this branch and compile using make build.
- Start hoverfly and capture the requests.
- Write python or ruby script that sends callback to your application.
- Register callback hook using command - ./hoverctl postserveactionhook --add testwebhook --binary python3  --script=<absolute path to callback script> --delay=<delay in milliseconds to simulate callback delay>
- Once, done, you can check all registered hooks using curl - curl -X GET "http://localhost:8888/api/v2/hoverfly/post-serve-actions"
- Edit simulation schema JSON and add a hook in the response part of the request/response pair. 
- Sample for hook to be added in simulation schema - "postSimulationHookName": "<hook name>"
- You can now switch to capture mode and see if a callback comes or not back to the server.

kapishmalik:
@tommysitu  whenever you get time, could you please check this PR.

petkostas:
@kapishmalik I am going to give it a try, most probably by the end of today! Massive thanks, mate!

core/handlers/v2/hoverfly_postserveactiondetails_handler.go
```diff
@@ -0,0 +1,72 @@
...
+	"net/http"
+
+	"github.com/SpectoLabs/hoverfly/core/handlers"
+	"github.com/codegangsta/negroni"
+	"github.com/go-zoo/bone"
+)
+
+type HoverflyPostServeActionDetails interface {
+	GetPostServeActionDetails() PostServeActionDetailsView
+	RegisterPostServeActionHook(string, string, string, int) error
```
tommysitu, line 14: Perhaps just call it `PostServeAction` instead of `PostServeActionHook`, to me the former is clear enough

core/handlers/v2/hoverfly_postserveactiondetails_handler.go
```diff
@@ -0,0 +1,72 @@
...
+	Hoverfly HoverflyPostServeActionDetails
+}
+
+func (postServeActionDetailsHandler *HoverflyPostServeActionDetailsHandler) RegisterRoutes(mux *bone.Mux, am *handlers.AuthHandler) {
+
+	mux.Get("/api/v2/hoverfly/post-serve-actions", negroni.New(
+		negroni.HandlerFunc(am.RequireTokenAuthentication),
+		negroni.HandlerFunc(postServeActionDetailsHandler.Get),
+	))
+	mux.Post("/api/v2/hoverfly/post-serve-actions/hook", negroni.New(
```
tommysitu, line 28: let's just simplify the path to `/api/v2/hoverfly/post-serve-actions`

tommysitu, line 28: how about making this an idempotent endpoint and use `PUT` instead, it's how the middleware endpoint works, either create or update

core/handlers/v2/postserveactiondetails_views.go
```diff
@@ -0,0 +1,12 @@
+package v2
+
+type PostServeActionDetailsView struct {
+	Hooks []HookView `json:"hooks,omitempty"`
+}
+
+type HookView struct {
```
tommysitu, line 7: prefer `ActionView` then `HookView`

core/handlers/v2/postserveactiondetails_views.go
```diff
@@ -0,0 +1,12 @@
...
+	HookName            string `json:"hookName"`
+	Binary              string `json:"binary"`
+	ScriptContent       string `json:"script"`
+	DelayInMilliSeconds int    `json:"delayInMilliSeconds,omitempty"`
```
tommysitu, line 11: let's keep field name concise like `delayInMs`

core/handlers/v2/simulation_views_v5.go
```diff
@@ -65,16 +65,17 @@ func NewMatcherView(matcher string, value interface{}) MatcherViewV5 {
...
+	Body                   string                 `json:"body"`
+	BodyFile               string                 `json:"bodyFile,omitempty"`
+	EncodedBody            bool                   `json:"encodedBody"`
+	Headers                map[string][]string    `json:"headers,omitempty"`
+	Templated              bool                   `json:"templated"`
+	TransitionsState       map[string]string      `json:"transitionsState,omitempty"`
+	RemovesState           []string               `json:"removesState,omitempty"`
+	FixedDelay             int                    `json:"fixedDelay,omitempty"`
+	LogNormalDelay         *LogNormalDelayOptions `json:"logNormalDelay,omitempty"`
+	PostSimulationHookName string                 `json:"postSimulationHookName,omitempty"`
```
tommysitu, line 78: `postServeAction` instead of `postSimulationHookName` ?

hoverctl/cmd/postserveactionhook.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/SpectoLabs/hoverfly/hoverctl/configuration"
+	"github.com/SpectoLabs/hoverfly/hoverctl/wrapper"
+	"github.com/spf13/cobra"
+)
+
+var binary, scriptPath, hookNameToBeAdded, hookNameToBeDeleted string
+var delayInMilliSeconds int
+
+var postServeActionHookCommand = &cobra.Command{
+	Use:   "postserveactionhook",
```
tommysitu, line 15: `hoverctl post-serve-action` is easier to read

tommysitu (commented):
very straight forward piece, mainly some suggestions on the namings

kapishmalik:
@tommysitu  I have incorporated review comment changes and added a few more things to PR. I will be adding UTs for new files and documentation as separate PR. I have tested it once locally. I would request you also to test it once before you merge this PR.

kapishmalik:
@tommysitu I have added a few things which I want to point out.

- I have added the request/response pair as stdin to post serve action script.
- post serve action will be invoked on simulate/spy/synthesis mode whenever response will be served by hoverfly
- Added GET, PUT and DELETE APIs for getting/setting or deleting post serve actions
- Added hoverctl command to get/set/delete post serve actions

petkostas:
I am missing the part for adding the script directly with hoverfly (e.g. `hoverfly --help`) instead of only `hoverctl`. This is important for me as most of my simulations are predefined upfront.
For example:
```
./hoverfly --import imposter.json
INFO[2023-02-22T20:02:37+01:00] Using memory backend
INFO[2023-02-22T20:02:37+01:00] payloads imported                             failed=0 successful=1 total=2
FATA[2023-02-22T20:02:37+01:00] Failed to import given resource               error="invalid post server action name provided" import=imposter.json
```

kapishmalik:
@petkostas we already have support to add script via post-serve-actions.
Go to target directory and type:

./hoverctl post-serve-actions --set "action name" --binary python --script="script-location" --delay=1000

kapishmalik:
@petkostas I can add that as separate PR. As of now, we just want to understand if this works for you.

kapishmalik:
@petkostas for now, you can use hoverctl to create post serve actions and then using hoverctl only import simulation schema so that your automation won't break and you will also be able to use this feature from now.

petkostas:
@kapishmalik I got the hoverctl working (importing the post-action hook), yet my problem is that I mainly use `hoverfly` as we use Hoverfly through Docker images. To test this through `hoverctl` requires further actions, which I cannot perform in the environment I am now testing (as you can see, Hoverfly crashes).
I will finalise the test on my system, as I cannot test this currently in my pipeline flow, and get back to you! indeed, it looks neat and is what we need.

kapishmalik:
@petkostas sure makes sense. I would definitely come up with PR to add support in hoverfly cli in some days after completing UTs in new PR.

kapishmalik:
@tommysitu I personally also felt that it will be difficult for anyone to give entire script content in json payload. I am thinking to take script location instead of content in json payload. I will make this change as part of this PR itself.

kapishmalik:
@tommysitu Shall I accept script path/location instead of script content in our APIs? It will be helpful. I can see in middleware APIs we are asking for script content so thats why I did that way in post-serve-actions.

tommysitu:
@kapishmalik the API is mainly used by `hoverctl` to post the content to hoverfly, so it's ok the way it is similar to middleware. I think most of the users would be like @petkostas using `hoverfly` command directly to read a file from a location.

kapishmalik:
@tommysitu yup, I believe. I will be working on that feature next post adding UTs  for newly added code.

kapishmalik:
@tommysitu @petkostas I have added support to import post serve action using hoverfly ctl. We can import multiple post serve action. Can you guys test and let me know if anything else we should add.

./hoverfly -post-serve-action "test-webhook python "file path" 3000" -post-serve-action "test-webhook1 python "file path" 5000"  -import simulation.json

core/action/action.go
```diff
@@ -0,0 +1,137 @@
...
+
+	log.WithFields(log.Fields{
+		"stdin": string(pairViewBytes),
+	}).Info("Delaying to execute post serve action")
+
+	//adding 200 ms to include some buffer for it to return response
+	time.Sleep(time.Duration(200+action.DelayInMs) * time.Millisecond)
+
+	actionCommand := exec.Command(action.Binary, action.Script.Name())
+	actionCommand.Stdin = bytes.NewReader(pairViewBytes)
```
tommysitu, line 122: Is this for the action to access data from the request-response pair?

hoverctl/cmd/postserveaction.go
```diff
@@ -0,0 +1,75 @@
...
+
+var postServeActionCommand = &cobra.Command{
+	Use:   "post-serve-action",
+	Short: "Get, Set & Delete Hoverfly PostServeAction",
+	Long: `
+Hoverfly PostServeAction can be set using the following flags: 
+	--set --binary --script --delay
+
+Hoverfly PostServeAction can be deleted by passing action name to be deleted:
+	--delete
```
tommysitu, line 23: it'd be better to be consistent with other hoverctl commands:
```
hoverctl post-serve-action set --name --binary --script --delay
```
and 
```
hoverctl post-serve-action delete --name
```

tommysitu:
I've tested both hoverctl and hoverfly commands, works really well!

core/action/action.go
kapishmalik: yup, it may require request/response details and make callback response.

hoverctl/cmd/postserveaction.go
kapishmalik: very much required. Thanks for this. incorporated.

kapishmalik:
@tommysitu I have incorporated review comment changes. You can review.

core/action/action.go
```diff
@@ -0,0 +1,137 @@
...
+	return scriptInfo, nil
+}
+
+func setBinary(action *Action, binary string) error {
+	if binary == "" {
+		action.Binary = ""
+		return nil
+	}
+	action.Binary = binary
+	return nil
```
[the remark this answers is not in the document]
kapishmalik: didn't notice this. Done

tommysitu (commented):
Nice one @kapishmalik. If you add some unit tests and fix the merge conflict, it would be good to go. Thanks!

kapishmalik:
@tommysitu I am planning to add UTs and documentation as part of the new PR as I have added lot of new code so wanted to cover all the files. I have fixed the merge conflict.  Let me know if we are good with this for now.

tommysitu (approved):
🚀

kapishmalik:
@petkostas feature is there on master for you to use with support to add post serve action via hoverfly ctl. Any feedback is welcome. Just FYI... documentation is yet to be updated.

petkostas:
Amazing work @kapishmalik thank you very much for all the effort!
I will surely test it as it's something I was missing!

Issue #1060, opened 2023-02-24
Not able to fetch version hoverfly v2 


sachin-raja:
When I run `go get github.com/SpectoLabs/hoverfly/v2 `

I get an error stating 

> go: module github.com/SpectoLabs/hoverfly@upgrade found (v1.4.0), but does not contain package github.com/SpectoLabs/hoverfly/v2

Seems like the package has a versioning issue. 

## Additional relevant information
1. Hoverfly version: v2.0.0

tommysitu:
sorry we released the one with a wrong tag, the latest should be `v1.5.0`. It's still backward compatible. 

If you installed it through brew before, you can `brew uninstall hoverfly` and then re-install it again to pick up `v1.5.0`

PR #1061, opened 2023-02-24
fix: update go path for major version update

joobus:
The upgrade to v2.0.0 was not done properly and now v2 is not importable by any other programs.

[See similar issue here](https://stackoverflow.com/questions/58066548/installing-a-go-module-fails-with-invalid-module-contains-a-go-mod-file-so-ma).

[See Go wiki here](https://github.com/golang/go/wiki/Modules#semantic-import-versioning)

After merging this PR, you will need to make a patch release (v2.0.1) so that this library is importable again.

My main use-case is that I have written middleware in Go for hoverfly, and have imported the middleware request/response structures into my middleware, but now can't upgrade to v2 because of this issue.

This PR addresses https://github.com/SpectoLabs/hoverfly/issues/1060


tommysitu:
Oh no, my bad. @joobus I'm thinking to remove the v2 release and re-do it as v1.5.0 instead, because it's still backwards compatible. What do you think?

joobus:
@tommysitu I would support that.  Sticking to [semantic versioning](https://semver.org/) is a good idea.

tommysitu:
thanks, I will fix the versioning issue.

tommysitu:
this is fixed now: https://github.com/SpectoLabs/hoverfly/releases/tag/v1.5.0

sorry for the confusion.

PR #1068, opened 2023-03-08
Optimise and Refactor code to marshal/unmarshal form matcher value

kapishmalik:
@tommysitu could you please review this PR. It fixes an issue with the form matcher value as you reported.

"body": [
{
"matcher": "form",
"value": {
"json": [
{
"Matcher": "exact",
"Value": "{\"key\":\"value\"}",
"Config": null,
"DoMatch": null
}
]
}
}
]
```


tommysitu (approved):
nice one!

Issue #1002, opened 2022-01-27
Eager caching not behaving as anticipated.


colts45s:
## Steps to reproduce the issue
1. Start Hoverfly in capture mode
2. Run traffic through the proxy and capture both request and responses
3. Export configuration.json
4. Stop Hoverfly
5. Start Hoverfly in Simulation mode (Found webserver and proxy to behave identically)
6. Import configuration
7. Check cache. 

### Observed result
**None**
```
Cache is null
```

## Expected result
Cache populated with the configuration data in an eager fashion rather than lazy loading

## Additional relevant information
1. Hoverfly version: 1.3.0
2. I found the same behavior if switching from capture to simulate in the same instance. To get a cached request response pair I would have to execute the call in capture, change to simulation, then re-run the request to populate the cache.

tommysitu:
@colts45s this is a bug actually, so what happens is that the eager caching function is only called when you run `hoverctl mode simulate`, but it's not executed when hoverfly starts 🤦 

The workaround is that after step 5 in your example, you call `hoverctl mode simulate` and it should force the cache preloading.

colts45s:
I attempted this work around and was unable to achieve eager caching, I have a valid configuration.json file and even after i explicitly set the mode to simulate, loading the configuration does not add any items to the cache.

tommysitu:
ok, the problem is this function checks eagerly cacheable simulation always return nil 😞 
https://github.com/SpectoLabs/hoverfly/blob/master/core/models/request_matcher.go#L207

tommysitu:
https://github.com/SpectoLabs/hoverfly/pull/1006 should fix the workaround at least

kapishmalik:
@colts45s raised PR to fix this issue.

kapishmalik:
@colts45s this is done. You can check and let us know incase of any concerns.

kapishmalik:
@tommysitu you can close this issue.

tommysitu:
This has been fixed since https://github.com/SpectoLabs/hoverfly/releases/tag/v1.5.2

PR #1071, opened 2023-03-12
bugfix - Eagerly load cache on running in simulate/spy mode

kapishmalik:
Resolves => #1002 
@tommysitu  could you please review.

bugfix - Eagerly load cache on running in simulate/spy mode using hoverfly cmd

### DIFF
--- a/core/matching/cache_matcher.go
+++ b/core/matching/cache_matcher.go
@@ -139,7 +139,7 @@ func (this *CacheMatcher) FlushCache() error {
 }
 
 func (this *CacheMatcher) PreloadCache(simulation *models.Simulation) error {
-	log.Info("Entered for preloading cache")
+
 	if this.RequestCache == nil {
 		return errors.NoCacheSetError()
 	}
@@ -151,6 +151,6 @@ func (this *CacheMatcher) PreloadCache(simulation *models.Simulation) error {
 		}
 	}
 
-	log.Info("Exiting after preloading cache")
+	log.Info("Cache preloaded")
 	return nil
 }
--- a/core/matching/cache_matcher.go
+++ b/core/matching/cache_matcher.go
@@ -143,14 +143,21 @@ func (this *CacheMatcher) PreloadCache(simulation *models.Simulation) error {
 	if this.RequestCache == nil {
 		return errors.NoCacheSetError()
 	}
+	cacheRequestCount := 0
 	for _, pair := range simulation.GetMatchingPairs() {
 
 		if requestDetails := pair.RequestMatcher.ToEagerlyCacheable(); requestDetails != nil {
 			pairCopy := pair
 			this.SaveRequestMatcherResponsePair(*requestDetails, &pairCopy, nil)
+			cacheRequestCount = cacheRequestCount + 1
 		}
 	}
 
-	log.Info("Cache preloaded")
+	if cacheRequestCount > 0 && cacheRequestCount == len(simulation.GetMatchingPairs()) {
+		log.Info("cache preloaded completely")
+	} else if cacheRequestCount > 0 {
+		log.Info("cache preloaded partially")
+	}
+
 	return nil
 }
--- a/core/matching/cache_matcher.go
+++ b/core/matching/cache_matcher.go
@@ -3,7 +3,7 @@ package matching
 import (
 	"github.com/SpectoLabs/hoverfly/core/cache"
 	"github.com/SpectoLabs/hoverfly/core/errors"
-	"github.com/SpectoLabs/hoverfly/core/handlers/v2"
+	v2 "github.com/SpectoLabs/hoverfly/core/handlers/v2"
 	"github.com/SpectoLabs/hoverfly/core/models"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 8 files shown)


tommysitu:
Hi @kapishmalik thank you for looking into this bug. Could I suggest a few improvements please?

1. It'd be useful to see if the cache has been preloaded or not right from the command line, so logging an info would be really really helpful. 
2.  We only preload cache if the simulation has exact matchers only, but since v1.5.0, capturing query params could result in an array matcher, so the preloading won't work any more if that's the case. It'd be good to get it fixed too.
3. When using `hoverctl`, only calling `hoverctl mode simulate` will preload, but not using `hoverctl import`. In order to bring hoverctl to parity with hoverfly commands, we should fix `hoverctl import` as well
4. Would be good to add some functional tests for cache pre-loading to prevent future regression.

I hope it's not too much to ask for!

kapishmalik:
Sure @tommysitu, I will incorporate all of these and get back.

tommysitu:
> Sure @tommysitu, I will incorporate all of these and get back.

You're star

core/matching/cache_matcher.go
```diff
@@ -150,5 +151,6 @@ func (this *CacheMatcher) PreloadCache(simulation *models.Simulation) error {
 		}
 	}
 
+	log.Info("Exiting after preloading cache")
```
tommysitu, line 154: Something like `Cache preloaded` is ok

core/matching/cache_matcher.go
```diff
@@ -139,6 +139,7 @@ func (this *CacheMatcher) FlushCache() error {
 }
 
 func (this *CacheMatcher) PreloadCache(simulation *models.Simulation) error {
+	log.Info("Entered for preloading cache")
```
tommysitu, line 142: not needed

kapishmalik: done

kapishmalik:
I need to test and add/fix UTs and FTs. I will inform here once done.

kapishmalik:
@tommysitu I started writing FTs but found one FT saying that the cache should be flushed when importing via API. So, should we incorporate point 3 i.e. to preload the cache on doing the hoverctl import command that eventually call PUT api. 

```It("should be flushed when importing via the API", func() {
		hoverfly.ImportSimulation(testdata.JsonPayload)
		cacheView := hoverfly.GetCache()

		Expect(cacheView.Cache).To(HaveLen(0))
	})

**Comment by kapishmalik** [comment]
@tommysitu it ll be great if you can help me with all scenarios when cache should be preloaded? I will document also so that expectations should be clear from hoverfly tool of the users.

**Comment by tommysitu** [comment]
right, maybe I've missed something, let's skip `3` for now. Thanks!

**Comment by kapishmalik** [comment]
@tommysitu I have incorporated all the comments/suggestions provided above. Could you please check this PR.

File: core/matching/cache_matcher.go [hunk 3]
```diff
@@ -150,5 +151,6 @@ func (this *CacheMatcher) PreloadCache(simulation *models.Simulation) error {
 		}
 	}
 
+	log.Info("Cache preloaded")
```
tommysitu, line 154: this log should go into the `if` block where the cache is actually pre-loaded

core/models/request_matcher.go
```diff
@@ -313,10 +313,15 @@ func (this RequestMatcher) ToEagerlyCacheable() *RequestDetails {
 	if this.Query != nil && len(*this.Query) > 0 {
 		for key, valueMatchers := range *this.Query {
 			for _, valueMatcher := range valueMatchers {
-				if valueMatcher.Matcher != matchers.Exact {
+				if valueMatcher.Matcher != matchers.Exact && !(valueMatcher.Matcher == matchers.Array && (valueMatcher.Config == nil || len(valueMatcher.Config) == 0)) {
 					return nil
 				}
-				query[key] = []string{valueMatcher.Value.(string)}
+				if valueMatcher.Matcher == matchers.Array {
+					query[key] = valueMatcher.Value.([]string)
```
tommysitu, line 320: this line throws an array when I import a simulation with query array matcher: 
```
panic: interface conversion: interface {} is []interface {}, not []string

goroutine 1 [running]:
github.com/SpectoLabs/hoverfly/core/models.RequestMatcher.ToEagerlyCacheable({{0x140003e4810, 0x1, 0x1}, {0x140003e4840, 0x1, 0x1}, {0x140003e4870, 0x1, 0x1}, {0x140003e48a0, ...}, ...})
        /Users/tommy.situ/repos/kapish/hoverfly/core/models/request_matcher.go:320 +0x958
github.com/SpectoLabs/hoverfly/core/matching.(*CacheMatcher).PreloadCache(0x140002efd60, 0x16dd5b5f2?)
        /Users/tommy.situ/repos/kapish/hoverfly/core/matching/cache_matcher.go:148 +0xdc
main.main()
        /Users/tommy.situ/repos/kapish/hoverfly/core/cmd/hoverfly/main.go:554 +0x2ee4

```

You can try capture a simulation using this:
`curl --proxy localhost:8500 'time.jsontest.com?token=1&token=2'`

core/matching/cache_matcher.go
kapishmalik: got it... done

core/models/request_matcher.go
kapishmalik: incorporated

tommysitu (approved):
noice!

PR #1073, opened 2023-04-10
fix: upgrade net package due to vulnerabilities

joobus:
There are 2 vulnerabilities in the golang.org/x/net package:

- https://nvd.nist.gov/vuln/detail/CVE-2022-41723
- https://nvd.nist.gov/vuln/detail/CVE-2022-41717

This PR upgrades that package.


tommysitu (approved):
Thank you

PR #1095, opened 2023-09-12
Bugfix - return base64 encoded string for binary data in request view

kapishmalik:
Bugfix - return base64 encoded string for binary data in request in Journal. This change will also return same request view to local/remote middleware and post serve action scripts.


tommysitu (approved):
👍

PR #1102, opened 2023-11-06
fix: upgrade net package due to vulnerability

joobus:
One of my work tools is reporting a vulnerability in the `golang/x/net` package.  Upgrading that dependency.


tommysitu (approved):
Bump `golang.org/x/` packages to for fixing security vulnerabilities, large diff due to vendor update, all tests passed, LGTM.

PR #1103, opened 2023-11-07
Fix invalid JSON on Requiring State in order to Match

Warfront1:
This addresses a small JSON formatting error in the documentation.


tommysitu (approved):
Thanks!

PR #1115, opened 2024-02-28
Add -log-no-quotes CLI flag to disable quoting/escaping of fields in logs

benediktwerner:
It's currently quite hard to read error messages on a failed match or similar in the logs because the whole message including newlines is quoted and escaped. Even more so if the request or response also contains JSON which then leads to double quoting. This also makes it impossible to directly copy parts of the error message to adjust simulations.

Since v1.6, Logrus has an option to disable quotes which also disables escaping of fields. Adding a CLI option to enable that seemed like the most sensible option to me.


tommysitu (approved):
👍  love it

Issue #1101, opened 2023-10-31
Support request matching for missing parameters


martin-tarjanyi:
**Is your feature request related to a problem? Please describe.**
I need to create different stub responses for an endpoint based on the presence of a query parameter. I'm not able to create a matcher which would activate only when a parameter is missing.

**Describe the solution you'd like**
Support for a matcher for absent parameters or some kind of negative matcher.

**Describe alternatives you've considered**
I see there was a similar issue in the past (https://github.com/SpectoLabs/hoverfly/issues/889) but that's not applicable for my use case.

kapishmalik:
@tommysitu  You can assign this to me. I will work on the same.

PR #1117, opened 2024-03-11
Add negation matcher to hoverfly

kapishmalik:
resolves => #1101


tommysitu:
Thanks, I will test it, is this how you use it? 
```
{
   "matcher": "negate", 
   "value": "something else"
}
```

kapishmalik:
Hey @tommysitu  yes, that is way we can test it. I have added integration tests as well for reference. Also, I will add documentation once approach looks fine.

tommysitu (approved):
It looks great 👍

PR #1123, opened 2024-04-03
Add support for remote post serve action

kapishmalik:
Incorporated following changes:

- Added support for remote post serve action
- Unit Tests
- Integration Tests
- Documentation

### DIFF
--- a/core/action/action.go
+++ b/core/action/action.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"os/exec"
 	"path"
@@ -19,10 +20,11 @@ import (
 type Action struct {
 	Binary    string
 	Script    *os.File
+	Remote    string
 	DelayInMs int
 }
 
-func NewAction(actionName, binary, scriptContent string, delayInMs int) (*Action, error) {
+func NewLocalAction(actionName, binary, scriptContent string, delayInMs int) (*Action, error) {
 
 	scriptInfo := &Action{}
 	if strings.TrimSpace(actionName) == "" {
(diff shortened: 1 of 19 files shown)


core/action/action.go
```diff
@@ -41,6 +43,19 @@ func NewAction(actionName, binary, scriptContent string, delayInMs int) (*Action
 	return scriptInfo, nil
 }
 
+func NewRemoteAction(actionName, host string, delayInMs int) (*Action, error) {
+
+	if strings.TrimSpace(actionName) == "" {
+		return nil, errors.New("empty action name passed")
+	}
+
+	if !hasHttpPrefix(host) {
```
tommysitu, line 52: Can we use the following to validate url pls? 
```
func validateURL(u string) bool {
    _, err := url.ParseRequestURI(u)
    return err == nil
}
```

tommysitu (approved):
Looks great 👍, only one comment. 

Also could we include the following in the next PR please? 
- Log a message saying a (local/remote) post-serve-action is set, I think we have done the same for middleware. 
- Update the Journal entry to indicate the request has triggered a callback, for example: 
```
"response": {
        "status": 200,
        "body": "Hello World",
        "encodedBody": false,
        "headers": {
          "Content-Length": [
            "11"
          ]
        }
      },
      "postServeAction": {
        "name": "outbound-http",
        "correlationId": "1ce7ea0c-7b0d-4cc5-9f7c-4a406b0e3c48",
        "invoked": "2024-04-07T10:51:25.428+08:00",
        "completed": "2024-04-07T10:51:26.428+08:00",
        "httpStatus": 200
      },
```
We need to include `correlationId` and `httpStatus` code for remote action only: 
`correlationId` - a randomly generated UUID passed as `X-CORRELATION-ID` header to the remote action endpoint, user can use it for e2e tracing.
'httpStatus` - the response status code from the remote action endpoint.

Thanks!

core/action/action.go
kapishmalik: incorporated

PR #1124, opened 2024-04-08
Add logging and journal entry for post serve action

kapishmalik:
Add logging and journal entry for post serve actions.

### DIFF
--- a/core/action/action.go
+++ b/core/action/action.go
@@ -134,6 +134,7 @@ func (action *Action) Execute(pair *models.RequestResponsePair, journalIDChannel
 	time.Sleep(time.Duration(200+action.DelayInMs) * time.Millisecond)
 
 	journalID := <-journalIDChannel
+	close(journalIDChannel)
 	log.Info("Journal ID received ", journalID)
 
 	//if it is remote callback
--- a/core/action/action.go
+++ b/core/action/action.go
@@ -134,7 +134,6 @@ func (action *Action) Execute(pair *models.RequestResponsePair, journalIDChannel
 	time.Sleep(time.Duration(200+action.DelayInMs) * time.Millisecond)
 
 	journalID := <-journalIDChannel
-	close(journalIDChannel)
 	log.Info("Journal ID received ", journalID)
 
 	//if it is remote callback
--- a/core/action/action.go
+++ b/core/action/action.go
@@ -133,9 +133,6 @@ func (action *Action) Execute(pair *models.RequestResponsePair, journalIDChannel
 	//adding 200 ms to include some buffer for it to return response
 	time.Sleep(time.Duration(200+action.DelayInMs) * time.Millisecond)
 
-	journalID := <-journalIDChannel
-	log.Info("Journal ID received ", journalID)
-
 	//if it is remote callback
 	if action.Remote != "" {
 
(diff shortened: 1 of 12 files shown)


core/action/action.go
```diff
@@ -131,6 +133,10 @@ func (action *Action) Execute(pair *models.RequestResponsePair) error {
 	//adding 200 ms to include some buffer for it to return response
 	time.Sleep(time.Duration(200+action.DelayInMs) * time.Millisecond)
 
+	journalID := <-journalIDChannel
```
tommysitu, line 136: this is gonna be blocking, and the delay timing could be wrong.

core/action/action.go
```diff
@@ -131,6 +133,10 @@ func (action *Action) Execute(pair *models.RequestResponsePair) error {
...
+	close(journalIDChannel)
```
tommysitu, line 137: it's not recommended by Go to close the channel by the receiver: 
```
// The close built-in function closes a channel, which must be either
// bidirectional or send-only. It should be executed only by the sender,
// never the receiver, and has the effect of shutting down the channel after
// the last sent value is received. 
```
I think that's because only the sender knows when it finishes sending data, and close the channel safely.

core/hoverfly.go
```diff
@@ -220,12 +220,15 @@ func (hf *Hoverfly) processRequest(req *http.Request) *http.Response {
 	}
 
 	if result.PostServeActionInputDetails != nil {
+
+		journalIDChannel := make(chan string, 1)
```
tommysitu, line 224: it's possible that this channel never get closed, because `go postServeAction.Execute` is wrapped in an if statement.

core/action/action.go
tommysitu: probably move this to the end of the execution?

core/hoverfly_service_test.go
```diff
@@ -1371,12 +1371,6 @@ func TestHoverfly_GetPostServeActions(t *testing.T) {
 
 	Expect(postServeActions).NotTo(BeNil())
 	Expect(postServeActions.Actions).To(HaveLen(2))
-	Expect(postServeActions.Actions[0].ActionName).To(Equal("test-local-callback"))
-	Expect(postServeActions.Actions[0].Binary).To(Equal("python3"))
-	Expect(postServeActions.Actions[0].DelayInMs).To(Equal(1900))
-	Expect(postServeActions.Actions[1].ActionName).To(Equal("test-remote-callback"))
-	Expect(postServeActions.Actions[1].Remote).To(Equal("http://localhost"))
-	Expect(postServeActions.Actions[1].DelayInMs).To(Equal(1800))
```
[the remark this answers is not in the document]
kapishmalik: reverted. Removed by mistake.

core/action/action.go
kapishmalik: It won't be block because sending to channel will be instantaneous in another routine. I have put delay of 200ms before so that it waits for response to reach first followed by callback in the cases where delay is set to 0 explicitly by the user.

core/action/action.go
kapishmalik: Yea, makes sense. I thought of doing it at receiver end because we are sending from multiple places(2 places in proxy.go and unit tests) so thought of doing at receiver end.

core/hoverfly.go
kapishmalik: yup, correct. Made the required changes.

core/action/action.go
tommysitu: but if the sender for whenever reason fails to send an id, the action goroutine is  gonna be waiting indefinitely, I think it's safer to move this receiver to the end of the action execution, and also set a timeout using the select statement, because the last thing we want to have is lots of stuck action goroutines :(

core/action/action.go
```diff
@@ -142,26 +147,39 @@ func (action *Action) Execute(pair *models.RequestResponsePair) error {
...
 			return nil
 		}
 		log.Info("Remote post serve action invoked successfully")
 		return nil
 	}
 
+	invokedTime := time.Now()
 	actionCommand := exec.Command(action.Binary, action.Script.Name())
 	actionCommand.Stdin = bytes.NewReader(pairViewBytes)
+	completionTime := time.Now()
```
tommysitu, line 179: the completion time for local action should be on L194

core/action/action.go
kapishmalik: done

tommysitu (approved):
👍

PR #1126, opened 2024-05-03
capture on miss request in spy mode

kapishmalik:
Feature: capture on miss request in spy mode

I will add documentation and tests in separate PR.


tommysitu (approved):
Replay of https://github.com/SpectoLabs/hoverfly/pull/1069, LGTM. Thank you!

PR #1141, opened 2024-08-15
Methods to retrieve CSV data as array or map

stuioco:
Hi!

I added three methods to make iterating over CSV data possible from templating.

- csvAsMap - returns an array of maps of string representing the data in the CSV which makes this.fieldname possible: within an each block
- csvMatchingRows - is the same thing but filters records based on a fieldvalue and field to match
- csvAsArray - just returns the whole thing as an array of arrays of string. 

I also changed the name of the previous parseCSV to fetchSingleFieldCSV (to be more descriptive given these new ones.)
I have not updated the documentation. Will do so once you've reviewed.

### DIFF
--- a/core/templating/template_helpers.go
+++ b/core/templating/template_helpers.go
@@ -212,7 +212,7 @@ func (t templateHelpers) faker(fakerType string) []reflect.Value {
 	return []reflect.Value{}
 }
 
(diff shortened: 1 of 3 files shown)


tommysitu (approved):
LGTM

PR #1143, opened 2024-08-19
Manipulate csv contents

stuioco:
Added 3 templating functions to manipulate csv content.

- csvCountRows - returns the number of rows in the data source
- csvDeleteRows - deletes rows with a matching provided field value
- csvAddRow - adds a new row from a provided array

Docs are also updated

### DIFF
--- a/core/templating/datasource.go
+++ b/core/templating/datasource.go
@@ -2,15 +2,18 @@ package templating
 
 import (
 	"encoding/csv"
+	"strings"
+	"sync"
+
 	v2 "github.com/SpectoLabs/hoverfly/core/handlers/v2"
 	log "github.com/sirupsen/logrus"
-	"strings"
 )
 
 type DataSource struct {
 	SourceType string
 	Name       string
 	Data       [][]string
+	mu         sync.Mutex
 }
 
 func NewCsvDataSource(fileName, fileContent string) (*DataSource, error) {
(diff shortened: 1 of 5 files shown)


tommysitu (commented):
Could you explain what the use case would be please? Templating function is not supposed for mutating hoverfly data, their main purpose is to generate dynamic response data.

stuioco:
Hi Tommy, the main reason to extend the csv capability from read only to read/write is that it gives a much simpler and more natural mechanism to simulate structured data persistence between calls than having to use journal indexing which can be confusing for complex processes.

kapishmalik:
It is quite unconventional to do so. We can make another api call to add or delete the data instead of displaying it using template function.

stuioco:
I agree it's unconventional, but not necessarily a bad idea. I'm struggling to see how using just api calls we can implement a sandbox environment that needs to simulate a mutable persistent back-end that supports CRUD operations on known data structures.

kapishmalik:
Can't we write middleware to update CSV and it shows that data instead of templating function. Can you help us understand your use case?

tommysitu (approved):
A small extension for csv templating functions only, with no other functional changes. It offers a simple way to manipulate CSV data sources in the templated response, worth considering for some CSV data source users.

PR #1146, opened 2024-08-25
SQL over CSV with tests and docs

stuioco:
Added support for SELECT, UPDATE, DELETE over CSV data sources in csvSQL template function.

Intention is to allow users to simulate a persistent CRUD datastore in a straightforward manner.

This in addition to template function earlier for simpler datasource manipulation.

Have included tests and documentation.

I am not sure how the documentation will render once compiled.

### DIFF
--- a/core/templating/datasource.go
+++ b/core/templating/datasource.go
@@ -27,7 +27,7 @@ func NewCsvDataSource(fileName, fileContent string) (*DataSource, error) {
 	return &DataSource{"csv", fileName, records, sync.Mutex{}}, nil
 }
 
-func (dataSource DataSource) GetDataSourceView() (v2.CSVDataSourceView, error) {
+func (dataSource *DataSource) GetDataSourceView() (v2.CSVDataSourceView, error) {
 
 	content, err := getData(dataSource)
 	if err != nil {
@@ -36,7 +36,7 @@ func (dataSource DataSource) GetDataSourceView() (v2.CSVDataSourceView, error) {
 	return v2.CSVDataSourceView{Name: dataSource.Name, Data: content}, nil
 }
 
-func getData(source DataSource) (string, error) {
+func getData(source *DataSource) (string, error) {
 
 	var csvData strings.Builder
 	csvWriter := csv.NewWriter(&csvData)
--- a/core/templating/datasource.go
+++ b/core/templating/datasource.go
@@ -53,3 +53,12 @@ func getData(source *DataSource) (string, error) {
 	}
 	return csvData.String(), nil
 }
+
+func dataSourceExists(dataSources map[string]*DataSource, name string) bool {
+	for _, dataSource := range dataSources {
+		if dataSource.Name == name {
+			return true
+		}
+	}
+	return false
+}
--- a/core/templating/datasource_query.go
+++ b/core/templating/datasource_query.go
@@ -4,15 +4,9 @@ import (
 	"errors"
 	"regexp"
 	"strings"
-)
 
-// A mock structure representing your CSV data
-// var data = [][]string{
-// 	{"id", "name", "age", "city"},
-// 	{"1", "John", "30", "New York"},
-// 	{"2", "Jane", "25", "Los Angeles"},
-// 	{"3", "Doe", "40", "New York"},
-// }
+	log "github.com/sirupsen/logrus"
+)
 
 // RowMap represents a single row in the result set
 type RowMap map[string]string
@@ -25,45 +19,102 @@ type Condition struct {
 }
 
 // SelectQuery represents a simple SQL-like SELECT query
+//
+//	type SelectQuery struct {
+//		Columns    []string
+//		Conditions []Condition
+//	}
 type SelectQuery struct {
-	Columns    []string
-	Conditions []Condition
+	Columns        []string
+	Conditions     []Condition
+	DataSourceName string
 }
 
 // ParseQuery parses a query string into a SelectQuery struct
-func ParseQuery(query string) (SelectQuery, error) {
+// func ParseQuery(query string, headers []string) (SelectQuery, error) {
+// 	query = strings.TrimSpace(query)
+
+// 	selectRegex := regexp.MustCompile(`(?i)^SELECT\s+(.+)\s+WHERE\s+(.+)$`)
+// 	matches := selectRegex.FindStringSubmatch(query)
+// 	if len(matches) != 3 {
+// 		return SelectQuery{}, errors.New("invalid query format")
+// 	}
+
+// 	columnsPart := matches[1]
+// 	wherePart := matches[2]
+
+// 	columns := parseColumns(columnsPart, headers)
+// 	conditions, err := parseConditions(wherePart)
+// 	if err != nil {
+// 		return SelectQuery{}, err
+// 	}
+
+//		return SelectQuery{
+//			Columns:    columns,
+//			Conditions: conditions,
+//		}, nil
+//	}
+func ParseQuery(query string, datasource map[string]*DataSource) (SelectQuery, error) {
 	query = strings.TrimSpace(query)
 
-	selectRegex := regexp.MustCompile(`(?i)^SELECT\s+(.+)\s+WHERE\s+(.+)$`)
+	// Regular expression to match SELECT, FROM, and WHERE clauses
+	selectRegex := regexp.MustCompile(`(?i)^SELECT\s+(.+)\s+FROM\s+(\w+)\s*(?:WHERE\s+(.+))?$`)
 	matches := selectRegex.FindStringSubmatch(query)
-	if len(matches) != 3 {
+	if len(matches) < 3 {
 		return SelectQuery{}, errors.New("invalid query format")
 	}
 
 	columnsPart := matches[1]
-	wherePart := matches[2]
+	dataSourceName := matches[2]
+	wherePart := ""
+	if len(matches) == 4 {
+		wherePart = matches[3]
+	}
 
+	// Assuming headers can be retrieved from the data source at a later stage
+	//headers := []string{} // This will be populated once the data source is retrieved
+	headers := datasource[dataSourceName].Data[0]
+	log.Debug("###################")
+	log.Debug(headers)
+	log.Debug("###################")
+	// Determine the columns to select
+	columns := parseColumns(columnsPart, headers)
+
+	// Parse the WHERE clause if present
+	var conditions []Condition
+	var err error
+	if wherePart != "" {
+		conditions, err = parseConditions(wherePart)
+		if err != nil {
+			return SelectQuery{}, err
+		}
+	}
+
+	return SelectQuery{
+		Columns:        columns,
+		Conditions:     conditions,
+		DataSourceName: dataSourceName,
+	}, nil
+}
+
+// parseColumns determines the columns to select based on the query part and headers
+func parseColumns(columnsPart string, headers []string) []string {
+	columnsPart = strings.TrimSpace(columnsPart)
+	if columnsPart == "*" {
+		return headers
+	}
 	columns := strings.Split(columnsPart, ",")
 	for i, column := range columns {
 		columns[i] = strings.TrimSpace(column)
 	}
-
-	conditions, err := parseConditions(wherePart)
-	if err != nil {
-		return SelectQuery{}, err
-	}
-
-	return SelectQuery{
-		Columns:    columns,
-		Conditions: conditions,
-	}, nil
+	return columns
 }
 
 // parseConditions parses the WHERE part of the query into a slice of Conditions
 func parseConditions(wherePart string) ([]Condition, error) {
 	conditions := []Condition{}
 
-	conditionRegex := regexp.MustCompile(`(\w+)\s*(==|!=)\s*'([^']*)'`)
+	conditionRegex := regexp.MustCompile(`(\w+)\s*(==|!=|<=|>=|<|>)\s*'([^']*)'`)
 	conditionMatches := conditionRegex.FindAllStringSubmatch(wherePart, -1)
 
 	if len(conditionMatches) == 0 {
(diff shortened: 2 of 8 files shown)


kapishmalik:
Cool. This is nice 👍

core/templating/datasource.go
```diff
@@ -53,3 +53,12 @@ func getData(source DataSource) (string, error) {
 	}
 	return csvData.String(), nil
 }
+
+func dataSourceExists(dataSources map[string]*DataSource, name string) bool {
```
tommysitu, line 57: For more OOP pattern, this function should be on `TemplateDataSource`, as its interface should include, get / set / list / contain data source, etcs.

core/templating/template_helpers.go
```diff
@@ -135,34 +135,130 @@ func (t templateHelpers) isBool(s string) bool {
...
+// 	return isLessThan(valueToCheck, maximumValue)
+// }
+
+// func (t templateHelpers) isBetween(valueToCheck, minimumValue, maximumValue string) bool {
+// 	return t.isGreaterThan(valueToCheck, minimumValue) && t.isLessThan(valueToCheck, maximumValue)
+// }
+
+//########################
+
+// Comparison function type that takes two float64 values and returns a boolean.
```
tommysitu, line 208: it's better to remove the code if we don't need it.

core/templating/template_helpers.go
```diff
@@ -398,6 +496,56 @@ func (t templateHelpers) csvCountRows(dataSourceName string) string {
 	return fmt.Sprintf("%d", numRows)
 }
 
+func (t templateHelpers) csvSQL(queryString string) []RowMap {
```
tommysitu, line 499: go prefer camel case, so `csvSqlQuery` is more idiomatic

core/templating/template_helpers.go
```diff
@@ -398,6 +496,56 @@ func (t templateHelpers) csvCountRows(dataSourceName string) string {
...
+	templateDataSources := t.TemplateDataSource.DataSources
+	// Parse the query string to get the SelectQuery
+	query, err := parseQuery(strings.TrimSpace(queryString), templateDataSources)
+	if err != nil {
+		// Debugging: Print the parsed query
+		fmt.Printf("Error parsing query:: %+v\n", err)
+		log.Debug("Error parsing query:", err)
```
tommysitu, line 506: we should just do `log.Error` here, using the proper logging framework is more consistency and provide the full logging control and context.

core/templating/template_helpers.go
```diff
@@ -398,6 +496,56 @@ func (t templateHelpers) csvCountRows(dataSourceName string) string {
...
+
+	source.mu.Lock()
+	defer source.mu.Unlock()
+
+	var results []RowMap
+	var execErr error
+
+	switch query.Type {
+	case "SELECT":
+		results = executeSelectQuery(source.Data, query) //This is not by reference as we are not changing the data
```
tommysitu, line 525: you can still pass by pointer here if you are not mutating the data.

core/templating/template_helpers.go
```diff
@@ -398,6 +496,56 @@ func (t templateHelpers) csvCountRows(dataSourceName string) string {
...
+	case "UPDATE":
+		execErr = executeUpdateQuery(&source.Data, query) //This must be by reference as we are not changing the data
+	case "DELETE":
+		execErr = executeDeleteQuery(&source.Data, query) //This must be by reference as we are not changing the data
```
tommysitu, line 529: sql update/ delete returns number of affected rows and perhaps we should replicate it, and log a message as well so that we know something has happened

core/templating/templating_test backup.txt
```diff
@@ -0,0 +1,934 @@
+package templating_test
```
tommysitu, line 1: do we need this file?

core/templating/templating_test.go
```diff
@@ -194,6 +198,80 @@ func Test_ApplyTemplate_CsvCountRowsMissingDataset(t *testing.T) {
 	Expect(template).To(Equal(``))
 }
 
+func Test_ApplyTemplate_CsvSQL_Select(t *testing.T) {
+	RegisterTestingT(t)
+
+	template, err := ApplyTemplate(&models.RequestDetails{}, make(map[string]string), `{{#each (csvSQL "SELECT name FROM test-csv WHERE id == '1'")}}{{this.name}}{{/each}}`)
```
tommysitu, line 204: `{{#each (csvSQL "SELECT name FROM test-csv WHERE id == '1'")}}` does it work in a simulation response field in which it would be wrapped by another double quotes?

vendor/github.com/SpectoLabs/raymond/helper.go
```diff
@@ -389,6 +390,57 @@ func eachHelper(context interface{}, options *Options) interface{} {
 	return result
 }
 
+func firstHelper(context interface{}, options *Options) interface{} {
```
tommysitu, line 393: change to `vendor` package will get wiped the next time we run `go mod vendor`, so we need this change committed to `https://github.com/SpectoLabs/raymond` and then update `go.mod` to include the latest version of our raymond folk.

core/templating/datasource_query.go
```diff
@@ -0,0 +1,331 @@
...
+			results = append(results, projectRow(rowMap, query.Columns))
+		}
+	}
+	return results
+}
+
+// ExecuteUpdateQuery executes an UPDATE query and modifies the data in-place
+func executeUpdateQuery(data *[][]string, query SQLStatement) error {
+	if len(*data) < 2 {
+		return errors.New("no data available to update")
```
tommysitu, line 224: SQL doesn't fail if no data to update, it will return 0 for no affected rows

core/templating/datasource_query.go
```diff
@@ -0,0 +1,331 @@
...
+		}
+	}
+	return nil
+}
+
+// ExecuteDeleteQuery executes a DELETE query and modifies the data in-place
+func executeDeleteQuery(data *[][]string, query SQLStatement) error {
+	if len(*data) < 2 {
+		return errors.New("no data available to delete")
+	}
```
tommysitu, line 248: this too, sql wouldn't fail in this scenario

core/templating/datasource_query.go
```diff
@@ -0,0 +1,331 @@
...
+
+	headers := (*data)[0]
+	conditions := query.Conditions
+	filteredData := [][]string{headers}
```
tommysitu, line 252: it's gonna require to double the memory to run the delete query, you can try delete the row from the slice in place, such as: 
```
// removeRow removes the row at index rowIndex from the matrix.
func removeRow(matrix [][]int, rowIndex int) [][]int {
    if rowIndex < 0 || rowIndex >= len(matrix) {
        // Return the matrix unchanged if the rowIndex is out of bounds
        return matrix
    }

    // Overwrite the row at rowIndex with the next rows
    copy(matrix[rowIndex:], matrix[rowIndex+1:])

    // Truncate the slice to remove the last row which is now duplicate
    return matrix[:len(matrix)-1]
}
```
Its space complexity is `O(1)` and time complexity id `O(n)`

core/templating/datasource_query.go
```diff
@@ -0,0 +1,331 @@
...
+func executeUpdateQuery(data *[][]string, query SQLStatement) error {
+	if len(*data) < 2 {
+		return errors.New("no data available to update")
+	}
+
+	headers := (*data)[0]
+	conditions := query.Conditions
+	setClauses := query.SetClauses
+	for i, row := range (*data)[1:] {
+		rowMap := mapRow(headers, row)
```
tommysitu, line 231: I started to think that it might be better to store the data as `[]map[string]string` rather than `[][]string`  if we are doing a lot of querying.

core/templating/datasource_query.go
```diff
@@ -0,0 +1,331 @@
...
+func parseColumns(columnsPart string, headers []string) []string {
+	columnsPart = strings.TrimSpace(columnsPart)
+	if columnsPart == "*" {
+		return headers
+	}
+	columns := strings.Split(columnsPart, ",")
+	for i, column := range columns {
+		columns[i] = strings.TrimSpace(column)
+	}
+	return columns
```
tommysitu, line 147: it's not validating if the column a valid header or not

core/templating/datasource.go
stuioco: Thanks, I will move it

core/templating/template_helpers.go
stuioco: Whoops, yes I will delete

core/templating/template_helpers.go
stuioco: I will change

core/templating/template_helpers.go
stuioco: Yes, I will clean up

core/templating/template_helpers.go
stuioco: Thanks, I will change

core/templating/template_helpers.go
stuioco: Yes, I will change it

core/templating/templating_test backup.txt
stuioco: Whoops, I will clean up

core/templating/templating_test.go
stuioco: I will test. Currently only tested as part of a bodyFile so unescaped

vendor/github.com/SpectoLabs/raymond/helper.go
stuioco: Thanks, I will do so

core/templating/datasource_query.go
stuioco: I will change

core/templating/datasource_query.go
stuioco: Will do

core/templating/templating_test.go
stuioco: Yes, tested and works

core/templating/datasource_query.go
stuioco: Ok will do

core/templating/datasource_query.go
stuioco: Agreed - might be worth doing

core/templating/datasource.go
stuioco: Done

core/templating/template_helpers.go
stuioco: Done

core/templating/template_helpers.go
stuioco: Done

core/templating/template_helpers.go
stuioco: Done

core/templating/template_helpers.go
stuioco: Done

core/templating/template_helpers.go
stuioco: Done

core/templating/templating_test backup.txt
stuioco: Done

vendor/github.com/SpectoLabs/raymond/helper.go
stuioco: This I haven't done yet - I will asap though

core/templating/datasource_query.go
stuioco: Done

core/templating/datasource_query.go
stuioco: Done

core/templating/datasource_query.go
stuioco: Done

core/templating/datasource_query.go
tommysitu: let's leave it for another PR.

core/templating/datasource_sql_over_csv.go
```diff
@@ -0,0 +1,375 @@
...
+
+// ExecuteUpdateQuery executes an UPDATE query and modifies the data in-place
+func executeSqlUpdateCommand(data *[][]string, query SQLStatement) []RowMap {
+	if len(*data) < 2 {
+		log.Error("no data available to update")
+		return []RowMap{
+			{
+				"rowsAffected": "0",
+			},
+		}
```
tommysitu, line 240: I was thinking of returning an integer only, `UPDATE` / `DELETE` in SQL doesn't return a result set, so there is no need to wrap it in the row record. This special `"rowsAffected": "0"` will get returned to the templated response too which I think it's unnecessary.

I was thinking that  `executeSqlUpdateCommand` returns an int, then the caller `csvSqlCommand` will log the counts, and return nil for the result set.

core/templating/template_helpers.go
stuioco: I have updated update and delete to log rows and return nil

core/templating/datasource_query.go
stuioco: I have updated update and delete to log rows and return nil

core/templating/datasource_query.go
stuioco: I have updated update and delete to log rows and return nil

core/templating/template_datasource.go
```diff
@@ -36,3 +37,15 @@ func (templateDataSource *TemplateDataSource) GetAllDataSources() map[string]*Da
 
 	return templateDataSource.DataSources
 }
+
+func dataSourceExists(templateDataSource *TemplateDataSource, name string) bool {
```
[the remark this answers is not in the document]
stuioco: Done

core/templating/datasource_sql_over_csv.go
stuioco: I have updated update and delete to log rows and return nil

vendor/github.com/SpectoLabs/raymond/helper.go
tommysitu: your change in raymond is merged, you can include it in the go mod here and do go mod vendor

stuioco: Have done so Tommy - thanks for guidance

tommysitu: you need to literally run `go mod vendor` on the project to bring in the change, but never mind I will do it on master.

tommysitu (approved):
new templating functions only, LGTM

PR #1149, opened 2024-08-28
Encapsulate templating data source and optimize locking

tommysitu:


core/templating/template_datasource.go
```diff
@@ -5,47 +5,45 @@ import (
...
+	defer t.rwMutex.RUnlock()
 
-	for _, dataSource := range templateDataSource.DataSources {
-		if dataSource.Name == name {
-			return true
-		}
-	}
-	return false
+	source, exits := t.dataSources[name]
+	return source, exits
```
tommysitu, line 48: support safe concurrent access using read lock

PR #1154, opened 2024-10-08
Capture delay along with response

### LINKED ISSUE #1153: Capture timing of requests
(I'm new to Hoverfly)

I think Hoverfly could be useful not just for unit testing, but performance testing.
In (some forms of) performance testing, I would like to stub out my application's connections to downstream webservices and return fixed responses, but with realistic (data-driven) delay.

Could `Capture` mode be enhanced to grab not only request/response, but also the time taken for the roundtrip? This data could (minimally) be read by the developer to better-inform "hardcoded" delays, or a step further, could feed directly into `Simulate` mode.

**Describe alternatives you've considered**
Manually researching and configuring delays for requests.

kapishmalik:
Resolves #1153

### DIFF
--- a/core/hoverfly_funcs.go
+++ b/core/hoverfly_funcs.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/SpectoLabs/hoverfly/core/errors"
 	v2 "github.com/SpectoLabs/hoverfly/core/handlers/v2"
@@ -19,19 +20,20 @@ import (
 )
 
 // DoRequest - performs request and returns response that should be returned to client and error
-func (hf *Hoverfly) DoRequest(request *http.Request) (*http.Response, error) {
+func (hf *Hoverfly) DoRequest(request *http.Request) (*http.Response, *time.Duration, error) {
 
 	// We can't have this set. And it only contains "/pkg/net/http/" anyway
 	request.RequestURI = ""
 
 	client, err := GetHttpClient(hf, request.Host)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
+	start := time.Now()
 	resp, err := client.Do(request)
 
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	resp.Header.Set("Hoverfly", "Was-Here")
@@ -40,8 +42,8 @@ func (hf *Hoverfly) DoRequest(request *http.Request) (*http.Response, error) {
 		resp.Header.Add("Hoverfly", "Forwarded")
 	}
 
-	return resp, nil
-
+	elapsed := time.Since(start)
+	return resp, &elapsed, nil
 }
 
 // GetResponse returns stored response from cache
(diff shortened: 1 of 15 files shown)


core/hoverfly_funcs.go
```diff
@@ -40,8 +42,8 @@ func (hf *Hoverfly) DoRequest(request *http.Request) (*http.Response, error) {
 		resp.Header.Add("Hoverfly", "Forwarded")
 	}
 
-	return resp, nil
-
+	elapsed := time.Since(start)
```
tommysitu, line 45: I think the intention is to measure the elapsed time of the outbound call, so moving this to L34 would be more accurate.

core/modes/diff_mode.go
```diff
@@ -74,9 +74,10 @@ func (this *DiffMode) Process(request *http.Request, details models.RequestDetai
 		respHeaders := util.GetResponseHeaders(actualResponse)
 
 		actualResponseDetails := &models.ResponseDetails{
-			Status:  actualResponse.StatusCode,
-			Body:    respBody,
-			Headers: respHeaders,
+			Status:     actualResponse.StatusCode,
+			Body:       respBody,
+			Headers:    respHeaders,
+			FixedDelay: int(duration.Milliseconds()),
```
tommysitu, line 80: I don't think we should record it for diff mode, we don't want to fail the diffing because of variation in the delay value.

core/modes/modify_mode.go
```diff
@@ -47,9 +48,10 @@ func (this ModifyMode) Process(request *http.Request, details models.RequestDeta
 	}
 
 	pair.Response = models.ResponseDetails{
-		Status:  resp.StatusCode,
-		Body:    string(bodyBytes),
-		Headers: resp.Header,
+		Status:     resp.StatusCode,
+		Body:       string(bodyBytes),
+		Headers:    resp.Header,
+		FixedDelay: int(duration.Milliseconds()),
```
tommysitu, line 54: not here either

core/modes/diff_mode.go
kapishmalik: done

core/modes/modify_mode.go
kapishmalik: done

core/modes/capture_mode.go
```diff
@@ -68,9 +69,10 @@ func (this CaptureMode) Process(request *http.Request, details models.RequestDet
 	respHeaders := util.GetResponseHeaders(response)
 
 	responseObj := &models.ResponseDetails{
-		Status:  response.StatusCode,
-		Body:    respBody,
-		Headers: respHeaders,
+		Status:     response.StatusCode,
+		Body:       respBody,
+		Headers:    respHeaders,
+		FixedDelay: int(duration.Milliseconds()),
```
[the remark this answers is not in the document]
kapishmalik: done. I was also thinking the same but then skipped.

core/hoverfly_funcs.go
kapishmalik: done

tommysitu (approved):
Looks great. Thank you for the contribution!

kapishmalik:
#hacktoberfest-accepted

PR #1156, opened 2024-11-29
bugfix - resolve issue #1155

### LINKED ISSUE #1155: Hoverfly templating issue with number greater than 9007199254740991
## Description of the bug
I am trying  to get data from request to the response via templating.While trying it is observed for numbers greater than 9007199254740991 in the request , the response gets changed .
**Example** : 
If I am passing this as the request body :
"Input": {
        "inputNo1": "100",
        **"inputNo2": "0009007199254740999",**
        "inputNo3": "001"
    }

**Output Response:** 
"Output": {
        "outputNo1": "100",
        **"outputNo2": "0009007199254741000",**
        "outputNo3": "1"
}

**Template used : "{{ Request.Body 'jsonpath' '$.Input.inputNo2' }}"**
Use the above template to extract the data from request .
Here the data is send as string and the template is also placed within "".
### Observed result
**Hoverfly error messages seen (If none, say none)**
``` None
(paste here)
```
**If possible, add screenshots to help explain your problem**

1. Hoverfly version:  
     | hoverctl | master-4270 |
     | hoverfly | v1.9.5 
2. Anything that might help us to diagnose the problem: Could it be due to floating-point precision limitations? What is the solution to it .

kapishmalik:
Resolves #1155

### DIFF
--- a/core/util/util.go
+++ b/core/util/util.go
@@ -399,14 +399,14 @@ func jsonPath(query, toMatch string) interface{} {
 		return ""
 	}
 
-	// Jsonpath library converts large int into a string with scientific notion, the following
-	// reverts that process to avoid mismatching when using the jsonpath result for csv data lookup
-	floatResult, err := strconv.ParseFloat(result, 64)
-	// if the string is a float and a whole number
-	if err == nil && floatResult == float64(int64(floatResult)) {
-		intResult := int(floatResult)
-		result = strconv.Itoa(intResult)
-	}
+	//// Jsonpath library converts large int into a string with scientific notion, the following
+	//// reverts that process to avoid mismatching when using the jsonpath result for csv data lookup
+	//floatResult, err := strconv.ParseFloat(result, 64)
+	//// if the string is a float and a whole number
+	//if err == nil && floatResult == float64(int64(floatResult)) {
+	//	intResult := int(floatResult)
+	//	result = strconv.Itoa(intResult)
+	//}
 
 	// convert to array data if applicable
 	var data interface{}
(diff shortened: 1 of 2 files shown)


core/util/util.go
```diff
@@ -399,14 +399,14 @@ func jsonPath(query, toMatch string) interface{} {
...
-		result = strconv.Itoa(intResult)
-	}
+	//// Jsonpath library converts large int into a string with scientific notion, the following
+	//// reverts that process to avoid mismatching when using the jsonpath result for csv data lookup
+	//floatResult, err := strconv.ParseFloat(result, 64)
+	//// if the string is a float and a whole number
+	//if err == nil && floatResult == float64(int64(floatResult)) {
+	//	intResult := int(floatResult)
+	//	result = strconv.Itoa(intResult)
+	//}
```
tommysitu, line 409: but this introduces another issue...

kapishmalik: This is not correct way to do things. I am not sure why we added this hack.

tommysitu: so what would be the correct way?

kapishmalik: @tommysitu  is that change still needed?

tommysitu: yes, it's needed. But commenting it out, several tests will fail, eg. `Test_ApplyTemplate_PutAndGetValue`

kapishmalik: @tommysitu I have added handling for it.

tommysitu (approved):
seems like a decent solution, thanks!

PR #1158, opened 2024-11-30
Bugfix resolve #1155

### LINKED ISSUE #1155: Hoverfly templating issue with number greater than 9007199254740991

kapishmalik:
Resolves #1155

### DIFF
--- a/core/util/util.go
+++ b/core/util/util.go
@@ -14,6 +14,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"io/ioutil"
 	"k8s.io/client-go/util/jsonpath"
+	"math/big"
 	"math/rand"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


tommysitu (approved):
Decent solution, thanks.

Issue #1169, opened 2025-01-20
Precision loss in JSON matching


VintaChlenix:
## Description of the bug
In the process of comparison of JSONs there is a bug caused by unmarshaling json to interface{}. By default numbers parse to float64. It has 53 bits of mantissa. This causes to precision loss.

### Real example:
Add a request json with bigint field to Hoverfly. For example {"id":1236546565217899832645}.
Then try to send a request with an id different from added by the last number. For example {"id":1236546565217899832640}.
Hoverfly will miss the difference.

### Simple example:
Write a testcase in json or json partial matcher with the same JSONs.

### Observed result

![Image](https://github.com/user-attachments/assets/fd7a58b4-41f5-4db6-a744-63601853d62c)

### Expected result

Matchers can find the difference between two JSONS with bigint fields.

## Additional relevant information
1. Hoverfly version: latest
2. Anything that might help us to diagnose the problem

VintaChlenix:
#1170

tommysitu:
Thanks for the quick fix. I will get a release out.

PR #1170, opened 2025-01-20
Bugfix issue #1169

VintaChlenix:
Fixed integer precision loss via matching JSONs. It is caused by unmarshaling JSON that contains bigints to type interface{}. Now decoder.UseNumber() is used.


tommysitu (approved):
nice

PR #1171, opened 2025-01-20
fix: update version to 1.10.8

chenrui333:
bump version to 1.10.8

relates to https://github.com/Homebrew/homebrew-core/pull/204969


tommysitu:
oh silly me, I'll fix forward.

PR #1183, opened 2025-03-25
update chat badge

marcoioco:


tommysitu (approved):

tommysitu:
we need to update the badge in README as well.

marcoioco:
> we need to update the badge in README as well.

ok thanks, I'll do that now

PR #1184, opened 2025-03-25
Update/readthedocs

marcoioco:
Replaced Gitter badge with Discussions badge for GitHub discussions.
Removed mailing group link as it is not used.


tommysitu (approved):
👍

PR #1220, opened 2026-04-10
Fix/relative url proxy handler

### LINKED ISSUE #774: Why would this be a bad request?
-bash-4.1$ export PATH=$PATH:~
-bash-4.1$ hoverctl status

+------------+----------+
| Hoverfly   | running  |
| Admin port |     8888 |
| Proxy port |     8500 |
| Proxy type | forward  |
| Mode       | capture  |
| Middleware | disabled |
+------------+----------+
-bash-4.1$ telnet localhost 8500
Trying ::1...
telnet: connect to address ::1: Connection refused
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
GET / HTTP/1.1
Host: thecure.com
User-Agent: Mozilla/5.0 (compatible; Yahoo! Slurp; http://help.yahoo.com/help/us/ysearch/slurp) .default/1538949339-2
Accept: */*

HTTP/1.1 500 Internal Server Error
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Sun, 07 Oct 2018 23:33:21 GMT
Content-Length: 64

This is a proxy server. Does not respond to non-proxy requests.
^]

telnet> quit
Connection closed.
-bash-4.1$ 

https://github.com/SpectoLabs/goproxy/blob/master/proxy.go#L108-L110

Why does the url has to be absolute? If a Host header is provided

lobovit:
## Problem

When an HTTP/1.1 client sends a request with a relative URL (path only, no scheme/host) and a `Host` header a valid pattern per RFC 7230 goproxy routes it to `NonproxyHandler`. The default implementation returns `500 "This is a proxy server"`, making Hoverfly appear broken for clients that rely on this behaviour (e.g. some HTTP libraries and CLI tools).

Reproduces with:
```
curl --proxy http://localhost:8500 http://example.com
```

## Fix

Override `NonproxyHandler` in `NewProxy` to reconstruct the absolute URL from the `Host` header and re-dispatch the request through the normal proxy pipeline:

- Returns `400 Bad Request` if the `Host` header is absent (HTTP/1.0 style request with no host information  cannot be forwarded).
- Returns `500` (original behaviour) if the `Host` port matches the proxy's own port, preventing infinite recursion when a client connects directly to Hoverfly instead of using it as a proxy.
- Otherwise sets `r.URL.Scheme = "http"` and `r.URL.Host = r.Host`, then calls `proxy.ServeHTTP` the same pattern already used by `NewWebserverProxy`.

## Testing

Three regression tests added to `core/proxy_test.go` using raw TCP (required because `http.DefaultClient` always produces absolute URLs and would bypass `NonproxyHandler`):

- `Test_NewProxy_ShouldReturn400ForRelativeURLWithNoHostHeader`
- `Test_NewProxy_ShouldHandleRelativeURLWithHostHeader`
- `Test_NewProxy_ShouldReturn500WhenRelativeURLHostIsProxyItself`

Closes #774
```

### DIFF
--- a/core/proxy.go
+++ b/core/proxy.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"regexp"
@@ -22,6 +23,8 @@ import (
 
 var ProxyAuthorizationHeader string
 
+const errProxyAuthRequired = "407 Proxy authentication required"
+
 // Creates goproxy.ProxyHttpServer and configures it to be used as a proxy for Hoverfly
 // goproxy is given handlers that use the Hoverfly request processing
 func NewProxy(hoverfly *Hoverfly) *goproxy.ProxyHttpServer {
@@ -29,6 +32,27 @@ func NewProxy(hoverfly *Hoverfly) *goproxy.ProxyHttpServer {
 
 	// creating proxy
 	proxy := goproxy.NewProxyHttpServer()
+
+	// Fix #774: HTTP/1.1 clients may send requests with a relative URL (path only) and a Host header.
+	// goproxy's default NonproxyHandler returns 500 for these. Reconstruct the absolute URL from the
+	// Host header so the request is processed normally, matching the behaviour of NewWebserverProxy.
+	//
+	// Guard: if the Host header targets the proxy's own port the client is connecting directly to
+	// Hoverfly (not using it as a proxy). In that case keep the original 500 "is a proxy server"
+	// response — forwarding to ourselves would cause infinite recursion or a panic.
+	proxy.NonproxyHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Host == "" {
+			http.Error(w, "Cannot handle requests without Host header, e.g., HTTP 1.0", http.StatusBadRequest)
+			return
+		}
+		if _, port, err := net.SplitHostPort(r.Host); err == nil && port == hoverfly.Cfg.ProxyPort {
+			http.Error(w, "This is a proxy server. Does not respond to non-proxy requests.", http.StatusInternalServerError)
+			return
+		}
+		r.URL.Scheme = "http"
+		r.URL.Host = r.Host
+		proxy.ServeHTTP(w, r)
+	})
 
 	if hoverfly.Cfg.AuthEnabled {
 		log.Info("Enabling proxy authentication")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tommysitu** [comment]
hi @scovl thanks for your contributions! Just let you know that I will be offline for the next few days in case you're wondering why no one picks up the review. 

no sure if @marcoioco is around, otherwise i will check these out as soon as I can!

**Comment by scovl** [comment]
> hi @scovl thanks for your contributions! Just let you know that I will be offline for the next few days in case you're wondering why no one picks up the review.
> 
> no sure if @marcoioco is around, otherwise i will check these out as soon as I can!

Np, dude!
Ty!

**Comment by marcoioco** [comment]
@scovl thanks. Will review this in a few days.

File: core/proxy.go [hunk 1]
```diff
@@ -30,6 +33,27 @@ func NewProxy(hoverfly *Hoverfly) *goproxy.ProxyHttpServer {
...
 	proxy := goproxy.NewProxyHttpServer()
 
+	// Fix #774: HTTP/1.1 clients may send requests with a relative URL (path only) and a Host header.
+	// goproxy's default NonproxyHandler returns 500 for these. Reconstruct the absolute URL from the
+	// Host header so the request is processed normally, matching the behaviour of NewWebserverProxy.
+	//
+	// Guard: if the Host header targets the proxy's own port the client is connecting directly to
+	// Hoverfly (not using it as a proxy). In that case keep the original 500 "is a proxy server"
+	// response — forwarding to ourselves would cause infinite recursion or a panic.
+	proxy.NonproxyHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
tommysitu, line 43: does it override the default goproxy implementation?

scovl: Yes, it overrides the default goproxy `NonproxyHandler`, which returns HTTP 500 for all non-proxy requests. The new handler reconstructs the absolute URL from the `Host` header and re-dispatches the request through `proxy.ServeHTTP`, enabling HTTP/1.1 clients that send relative URLs to be handled correctly. 

> The original 500 behavior is preserved only when the `Host` targets the proxy's own port, to prevent infinite recursion.

tommysitu: Great! I think this should fix the issue while improving the existing non-proxy request check.

tommysitu (approved):
Nice fix — the approach is sound and the placement right after goproxy.NewProxyHttpServer() is correct, since it overrides the default NonproxyHandler before any other handlers are wired.

RFC compliance. HTTP/1.1 mandates a Host header (RFC 7230 §5.4), so the 400 for missing Host is correct. HTTP/1.0 clients without Host genuinely can't be reconstructed — 400 is the only sane response there too.

HTTPS scope. Hard-coding scheme = "http" is fine — raw TCP clients like telnet can't do TLS anyway, and HTTPS on the proxy port requires CONNECT. 

Good test coverage too. Well done!

PR #1229, opened 2026-05-28
feat: add capture-on-miss-flag

schthms:
Adds a -capture-on-miss CLI flag that, when combined with -spy, configures spy mode to save request/response pairs to the simulation store on cache misses. Previously this behavior was only configurable via the admin API after startup.

Also adds unit tests covering the CaptureOnMiss path in spy mode and the SetModeWithArguments wiring for the new argument.

### DIFF
--- a/core/cmd/hoverfly/main.go
+++ b/core/cmd/hoverfly/main.go
@@ -38,6 +38,7 @@ import (
 	hvc "github.com/SpectoLabs/hoverfly/core/certs"
 	cs "github.com/SpectoLabs/hoverfly/core/cors"
 	"github.com/SpectoLabs/hoverfly/core/handlers"
+	v2 "github.com/SpectoLabs/hoverfly/core/handlers/v2"
 	"github.com/SpectoLabs/hoverfly/core/matching"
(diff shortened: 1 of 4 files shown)


tommysitu:
@claude

tommysitu:
Hi @schthms, the change is sensible, just some minor issues spotted by claude which you might want to address as well.

schthms:
Thanks @tommysitu. I addressed the review comments:
* Moved the -capture-on-miss wiring to after all imports complete. SetModeWithArguments calls PreloadCache internally, so placing it before imports meant the cache was built against an empty simulation when using HOVERFLY_IMPORT_RECORDS. The simulation is now fully loaded before the cache is preloaded in all cases.
* Added a fatal error when -capture-on-miss is passed without -spy so it fails loudly instead of silently doing nothing.
* Strengthened the CaptureOnMiss unit test with an assertion on savedArguments to verify the arguments actually reach the Save call.
* Added functional tests for the capture-on-miss path: one verifying a cache miss saves the pair, one verifying a cache hit does not.

tommysitu (approved):
Adding new flag to hoverfly cli, no-ops for existing feature. LGTM

PR #1243, opened 2026-09-02
docs: replace dead jsontest.com example URLs

spramanigit:
`jsontest.com` (used as the example host in the delays and specific-URLs tutorials) is no longer live - it now returns a Cloudflare 530 (origin unreachable) error, so these tutorials no longer work as written for anyone following along.

This replaces the dead URLs with two live public test APIs: `api.ipify.org` in place of `ip.jsontest.com` / `time.jsontest.com` (kept the "contains ip" property the specific-URLs tutorial's destination filter relies on), and `httpbin.org` in place of `date.jsontest.com` / the second delay-example host.

Touches both tutorials end to end: the .rst prose, the shell script examples, and the shared multiple-hosts-delay-simulation.json fixture used by the delays tutorial. Happy to adjust to a different replacement host if you'd prefer - just wanted to get the dead link fixed since it breaks the copy-paste examples as they stand.

### DIFF
--- a/docs/pages/simulations/multiple-hosts-delay-simulation.json
+++ b/docs/pages/simulations/multiple-hosts-delay-simulation.json
@@ -26,12 +26,12 @@
       "globalActions": {
          "delays": [
             {
-               "urlPattern": "time\\.jsontest\\.com",
+                "urlPattern": "api\\.ipify\\.org",
                "httpMethod": "",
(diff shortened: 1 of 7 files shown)


tommysitu (approved):
cool, I will revert https back to http because that will break the tutorial.

